PR #100, opened 2015-02-18
Fix print function

BenPope:


msotclas:
Hi __@BenPope__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, MSOTBOT;

kirkshoop:
this turns compile-time iteration into runtime recursion (that the compiler might collapse). I prefer the compile-time iteration because I have learned to prefer iteration over recursion. A more idiomatic way to express compile-time left-to-right iteration of a pack is welcome, it would be useful in some of the operators as well.
Is there a reason why you prefer runtime recursion?

BenPope:
I guess I just assumed it would all be inlined.  I was trying to silence a warning (on the line that says "silence warning"!  Yours is a neat trick, and I can't get the compiler to emit a warning now, so move along, nothing to see here.

PR #130, opened 2015-05-17
fix bug reported in test scheduler

kirkshoop:
Thanks to @d-led for reporting this issue.  (https://github.com/d-led/ticker/commit/70dec6397e66bf8deb57cefc65fd751efb9453ff#commitcomment-11227516)

### DIFF
--- a/Rx/v2/src/rxcpp/schedulers/rx-virtualtime.hpp
+++ b/Rx/v2/src/rxcpp/schedulers/rx-virtualtime.hpp
@@ -82,10 +82,8 @@ public:
                     }
                     next.what(r.get_recurse());
                 }
-                else {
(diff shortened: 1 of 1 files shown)


d-led:
great! looking forward to update and blog. Was my use of the test scheduler as it's intended?

kirkshoop:
@d-led - No that usage is not intended, advance_by is a private method right now. I am thinking that it should be made available.
The tests in the rxcpp repo show the intended usage of the test scheduler.

d-led:
will take a closer look! As for the API, I was thinking of something like [*that](https://github.com/d-led/reactiveexamples/blob/master/4%20newviewmodels.test/BackgroundTickerTests.cs#L20-L32).

PR #131, opened 2015-05-17
expose advance and sleep methods

kirkshoop:
add helpers for making coordinations.

### DIFF
--- a/Rx/v2/src/rxcpp/rx-coordination.hpp
+++ b/Rx/v2/src/rxcpp/rx-coordination.hpp
@@ -165,6 +165,11 @@ inline identity_one_worker identity_current_thread() {
     return r;
 }
 
+inline identity_one_worker identity_same_worker(rxsc::worker w) {
+    static identity_one_worker r(rxsc::make_same_worker(w));
+    return r;
+}
+
 class serialize_one_worker : public coordination_base
 {
     rxsc::scheduler factory;
@@ -291,6 +296,10 @@ inline serialize_one_worker serialize_new_thread() {
     return r;
 }
 
+inline serialize_one_worker serialize_same_worker(rxsc::worker w) {
+    static serialize_one_worker r(rxsc::make_same_worker(w));
+    return r;
+}
 
 }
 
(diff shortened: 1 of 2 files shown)


kirkshoop:
@d-led, this will allow code like the following

```
SCENARIO("showing off the test scheduler") {

    GIVEN("a test scheduler") {
        auto sc = rxcpp::schedulers::make_test();
        auto worker = sc.create_worker();
        auto test = rxcpp::identity_same_worker(worker);

        int count = 0;

        WHEN("subscribe to an observable sequence on the scheduler") {
            auto seq = rxcpp::observable<>::interval(
                std::chrono::milliseconds(1),
                test // on the test scheduler
            ).filter([](int i) { return i % 2; });

            seq.subscribe([&count](int){
                count++;
            });

            THEN("the sequence is not run at first") {
                std::this_thread::sleep_for(std::chrono::milliseconds(2));

                CHECK(count == 0);

                AND_WHEN("the test scheduler is advanced manually") {

                    THEN("the sequence is run as expected") {
                        worker.advance_by(10 /* ms */);
                        CHECK(count == 5);

                        AND_THEN("the scheduler can be advanced again") {
                            worker.advance_by(10);
                            CHECK(count == 10);
                        }
                    }
                }
            }
        }
    }

}
```

kirkshoop:
Although `std::this_thread::sleep_for(std::chrono::milliseconds(2));` is not very illustrative since the test scheduler only runs when advanced or started.

The THEN sections should be more like this

```
            THEN("the sequence is not run at first") {
                worker.sleep(2 /* ms */);
                CHECK(count == 0);

                AND_WHEN("the test scheduler is advanced manually") {

                    THEN("the sequence is run as expected") {
                        worker.advance_by(8 /* ms */);
                        CHECK(count == 5);

                        AND_THEN("the scheduler can be advanced again") {
                            worker.advance_by(10);
                            CHECK(count == 10);
                        }
                    }
                }
            }
```

d-led:
superb!

PR #141, opened 2015-06-04
Reduce

ValeryKopylov:
Implement new behavior of reduce operator:
1. Reducing an empty observable returns result_selector(seed).
2. Exceptions from accumulator() and result_selector() are caught and emitted as on_error().
3. first(), last(), sum(), average() emit rxcpp::empty_error when applied to an empty observable.
4. count() of an empty observable emits 0.
5. If the source observable for count(), sum(), average() calls on_error(), the resulting observable behaves like it was on_completed().

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-reduce.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-reduce.hpp
@@ -9,6 +9,14 @@
 
 namespace rxcpp {
 
+class empty_error: public std::runtime_error
+{
+    public:
+        empty_error(const std::string& msg):
+            std::runtime_error(msg)
+        {}
+};
+
 namespace operators {
 
 namespace detail {
@@ -289,14 +297,6 @@ auto reduce(Seed s, Accumulator a, ResultSelector rs)
 }
 
 }
-
-class empty_error: public std::runtime_error
-{
-    public:
-        empty_error(const std::string& msg):
-            std::runtime_error(msg)
-        {}
-};
 
 }
 
--- a/Rx/v2/src/rxcpp/operators/rx-reduce.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-reduce.hpp
@@ -12,7 +12,7 @@ namespace rxcpp {
 class empty_error: public std::runtime_error
 {
     public:
-        empty_error(const std::string& msg):
+        explicit empty_error(const std::string& msg):
             std::runtime_error(msg)
         {}
 };
@@ -91,19 +91,17 @@ struct reduce : public operator_base<rxu::value_type_t<reduce_traits<T, SourceOp
         ~reduce_initial_type()
         {
         }
-        reduce_initial_type(source_type o, accumulator_type a, result_selector_type rs, seed_type s, bool _abort_on_error)
+        reduce_initial_type(source_type o, accumulator_type a, result_selector_type rs, seed_type s)
             : source(std::move(o))
             , accumulator(std::move(a))
             , result_selector(std::move(rs))
             , seed(std::move(s))
-            , abort_on_error(_abort_on_error)
         {
         }
         source_type source;
         accumulator_type accumulator;
         result_selector_type result_selector;
         seed_type seed;
-        bool abort_on_error;
 
     private:
         reduce_initial_type& operator=(reduce_initial_type o) RXCPP_DELETE;
@@ -113,8 +111,8 @@ struct reduce : public operator_base<rxu::value_type_t<reduce_traits<T, SourceOp
     ~reduce()
     {
     }
-    reduce(source_type o, accumulator_type a, result_selector_type rs, seed_type s, bool abort_on_error = true)
-        : initial(std::move(o), std::move(a), std::move(rs), std::move(s), abort_on_error)
+    reduce(source_type o, accumulator_type a, result_selector_type rs, seed_type s)
+        : initial(std::move(o), std::move(a), std::move(rs), std::move(s))
     {
     }
     template<class Subscriber>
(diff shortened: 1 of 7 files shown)


Rx/v2/src/rxcpp/operators/rx-reduce.hpp
```diff
@@ -128,23 +138,39 @@ struct reduce : public operator_base<rxu::value_type_t<reduce_traits<T, SourceOp
...
+            state->out.on_next(result.get());
+            state->out.on_completed();
+        };
         state->source.subscribe(
             state->out,
         // on_next
             [state](T t) {
-                auto next = state->accumulator(state->current, t);
-                state->current = next;
+                try {
```
kirkshoop: This try/catch should not be needed. The subscriber will call this on_next method under protection, if it throws, it will call on_error with the exception.

Rx/v2/src/rxcpp/operators/rx-reduce.hpp
```diff
@@ -128,23 +138,39 @@ struct reduce : public operator_base<rxu::value_type_t<reduce_traits<T, SourceOp
...
+                    state->current = next;
+                } catch (const std::exception&) {
+                    state->out.on_error(std::current_exception());
+                }
             },
         // on_error
-            [state](std::exception_ptr e) {
-                state->out.on_error(e);
+            [state, on_completed](std::exception_ptr e) {
+                if (state->abort_on_error) {
```
kirkshoop: reduce should never filter out errors from the source. Always pass e to the out.on_error();

Rx/v2/src/rxcpp/operators/rx-reduce.hpp
```diff
@@ -128,23 +138,39 @@ struct reduce : public operator_base<rxu::value_type_t<reduce_traits<T, SourceOp
             reduce_state_type& operator=(reduce_state_type o) RXCPP_DELETE;
         };
         auto state = std::make_shared<reduce_state_type>(initial, std::move(o));
+        auto on_completed = [state]() {
+            rxu::maybe<value_type> result;
+            try {
```
kirkshoop: rxcpp::util has an on_exception function that should be used instead of a manual try/catch whenever possible.

Rx/v2/src/rxcpp/operators/rx-reduce.hpp
```diff
@@ -230,7 +256,35 @@ struct average {
...
             return avg;
         }
-        return a.value;
+        throw rxcpp::empty_error("No elements");
+    }
+};
+
+template<class T>
+struct sum {
+    struct seed_type
```
kirkshoop: I think that seed_type should be maybe<T>.

Rx/v2/src/rxcpp/operators/rx-reduce.hpp
```diff
@@ -230,7 +256,35 @@ struct average {
...
+        {
+        }
+        T value;
+        bool has_accumulation;
+    };
+    seed_type seed() {
+        return seed_type{};
+    }
+    seed_type operator()(seed_type a, T v) {
+        a.value += v;
```
kirkshoop: Accumulate should just reset to the initial value (when the seed is empty) and only add when the seed is !empty. Also, use + not +=. sum should work for any type that has defined operator+() even if they have not defined operator+=().

Rx/v2/src/rxcpp/rx-observable.hpp
```diff
@@ -263,27 +263,107 @@ class blocking_observable
...
+
+        When the source observable is empty:
+        \snippet blocking_observable.cpp blocking last empty sample
+        \snippet output.txt blocking last empty sample
+    */
     T last() const {
         rxu::maybe<T> result;
-        subscribe([&](T v){result.reset(v);});
-        if (result.empty()) abort();
+        rxu::maybe<std::exception_ptr> error;
```
kirkshoop: I am putting the rethrow logic into blocking_subscribe, it should not be here.

Rx/v2/src/rxcpp/rx-observable.hpp
```diff
@@ -263,27 +263,107 @@ class blocking_observable
...
+        When the source observable emits at least one item:
+        \snippet blocking_observable.cpp blocking first sample
+        \snippet output.txt blocking first sample
+
+        When the source observable is empty:
+        \snippet blocking_observable.cpp blocking first empty sample
+        \snippet output.txt blocking first empty sample
+    */
     T first() {
-        rxu::maybe<T> result;
```
kirkshoop: I think this change to first() should be reverted. the old code is a lot more efficient than using the non-blocking operator. just adding
```cpp
        if (result.empty())
            throw rxcpp::empty_error("No elements");
```
as was done in last() will cover it since blocking_subscribe will rethrow the error.

Rx/v2/src/rxcpp/rx-observable.hpp
```diff
@@ -263,27 +263,107 @@ class blocking_observable
...
-        subscribe([&](T v){result.reset(v);});
-        if (result.empty()) abort();
+        rxu::maybe<std::exception_ptr> error;
+        subscribe(
+            [&](T v){result.reset(v);},
+            [&](std::exception_ptr ep){error.reset(ep);});
+        if (!error.empty())
+            std::rethrow_exception(error.get());
+        if (result.empty())
+            throw rxcpp::empty_error("No elements");
```
kirkshoop: I think that the string passed to empty_error should not just say the same thing as the name 'empty_error'.
I would like something like this
`rxcpp::empty_error("last() requires a stream with at least one value")`

Rx/v2/src/rxcpp/operators/rx-reduce.hpp
```diff
@@ -9,6 +9,14 @@
 
 namespace rxcpp {
 
+class empty_error: public std::runtime_error
+{
+    public:
+        empty_error(const std::string& msg):
```
kirkshoop: I think that this constructor should be marked `explicit`.

kirkshoop:
I pushed these commits with a few changes. Thanks!

Issue #151, opened 2015-06-19
Observing on main thread


hasyimibhar:
In RxAndroid, there's `.observeOn(AndroidSchedulers.mainThread())`. How do you achieve that in RxCpp?

For example, how do I ensure that the call to `printf` below is executed on the main thread?
```cpp
auto numbers = rxcpp::observable<>::range(1, 100).subscribe_on(rxcpp::observe_on_new_thread());
numbers
    .map([](int n) { return n * 2; })
    .map([](int n) { return n + 1; })
    .as_blocking()
    .subscribe([](int n)
    {
        printf("Number: %d\n", n);
    });

**Comment by kirkshoop** [comment]
C++ does not have a standard UI thread. Each UI toolkit provides its own UI thread and a way to move work onto the UI Thread. Thus rxcpp has no built-in main thread scheduler.

However, I had been planning to write a scheduler that could be easily integrated with various UI Thread implementations.

Thanks to your question I now have a PR (#154) with a run-loop scheduler that allows expressions to observe results on the thread that the run_loop owns. 

## Example

```cpp
#include "rxcpp/rx.hpp"
// create alias' to simplify code
// these are owned by the user so that
// conflicts can be managed by the user.
namespace rx=rxcpp;

int main()
{
    std::cout << std::setw(16) << std::this_thread::get_id() << " <- main thread id" << std::endl;

    using namespace std::chrono;

    rx::schedulers::run_loop rl;

    auto mainthread = rx::observe_on_run_loop(rl);
    auto workthread = rx::synchronize_new_thread();

    rx::composite_subscription lifetime;

    rx::observable<>::interval(workthread.now() + milliseconds(100), milliseconds(400)).
        map([](int i){
            std::cout << std::setw(16) << std::this_thread::get_id() << ": " << i << std::endl;
            return i;
        }).
        take_until(rx::observable<>::timer(seconds(1))).
        subscribe_on(workthread).
        observe_on(mainthread).
        subscribe(lifetime, [&](int i){
            std::cout << std::setw(16) << std::this_thread::get_id() << ": " << i << std::endl;
        });

    while (lifetime.is_subscribed() || !rl.empty()) {
        while (!rl.empty() && rl.peek().when < rl.now()) {
            rl.dispatch();
        }
    }
    return 0;
}
```

## Output

```sh
  0x7fff7b1cb300 <- main thread id
     0x102047000: 1
  0x7fff7b1cb300: 1
     0x102047000: 2
[...]

Issue #153, opened 2015-06-23
memory leak when using repeat(void)


hoffis:
after running snippet below memory increase infinitely
```cpp
#include <rxcpp/rx.hpp>
#include <iostream>

int main(int argc,char** argv)
{
    rxcpp::sources::range(1,10)
    .repeat()
    .subscribe([](auto i)
    {
        std::cout << i << std::endl;
    });
    return 0;
}

```

kirkshoop:
Thanks! This is fixed.

PR #154, opened 2015-06-24
add run loop scheduler

kirkshoop:
I added the run loop scheduler to answer #151 

## Usage

```cpp
#include "rxcpp/rx.hpp"
// create alias' to simplify code
// these are owned by the user so that
// conflicts can be managed by the user.
namespace rx=rxcpp;

int main()
{
    std::cout << std::setw(16) << std::this_thread::get_id() << " <- main thread id" << std::endl;

    using namespace std::chrono;

    rx::schedulers::run_loop rl;

    auto mainthread = rx::observe_on_run_loop(rl);
    auto workthread = rx::synchronize_new_thread();

    rx::composite_subscription lifetime;

    rx::observable<>::interval(workthread.now() + milliseconds(100), milliseconds(400)).
        map([](int i){
            std::cout << std::setw(16) << std::this_thread::get_id() << ": " << i << std::endl;
            return i;
        }).
        take_until(rx::observable<>::timer(seconds(1))).
        subscribe_on(workthread).
        observe_on(mainthread).
        subscribe(lifetime, [&](int i){
            std::cout << std::setw(16) << std::this_thread::get_id() << ": " << i << std::endl;
        });

    while (lifetime.is_subscribed() || !rl.empty()) {
        while (!rl.empty() && rl.peek().when < rl.now()) {
            rl.dispatch();
        }
    }
    return 0;
}
```

## Output:

```sh
  0x7fff7b1cb300 <- main thread id
     0x102047000: 1
  0x7fff7b1cb300: 1
     0x102047000: 2
  0x7fff7b1cb300: 2
     0x102047000: 3
  0x7fff7b1cb300: 3
```

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-observe_on.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-observe_on.hpp
@@ -68,6 +68,22 @@ struct observe_on
                 , destination(std::move(d))
             {
             }
+            
+            void finish(std::unique_lock<std::mutex>& guard, typename mode::type end) const {
+                if (!guard.owns_lock()) {
+                    abort();
+                }
+                if (current == mode::Errored || current == mode::Disposed) {return;}
+                current = end;
+                queue_type fill_expired;
+                swap(fill_expired, fill_queue);
+                queue_type drain_expired;
+                swap(drain_expired, drain_queue);
+                RXCPP_UNWIND_AUTO([&](){guard.lock();});
+                guard.unlock();
+                lifetime.unsubscribe();
+                destination.unsubscribe();
+            }
 
             void ensure_processing(std::unique_lock<std::mutex>& guard) const {
                 if (!guard.owns_lock()) {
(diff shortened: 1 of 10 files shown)


kirkshoop:
Added fix for #153

b52:
@kirkshoop Quick question, how would you integrate your runloop with an existing runloop like e.g. Qt's. I'm trying to bind my observables to a `QObject` in a safe manner, i.e., using `observe_on`, but I'm unsure how to use it properly. Any hints or recommendations?

As a side note, it might make sense to aggregate such common problems in a document or wiki. I for myself like to collect all common problems and their solutions in the main readme file directly, this motives the usage better.

kirkshoop:
Hi!

The runloop scheduler is intended to allow the scheduled items to be dispatched in a loop - perhaps in the same loop with another event system - like Win32 window messages.

After some research QT appears to keep its event loop hidden inside the application object.
This means that runloop is not the best tool for the job.

It would not be terribly hard to build a scheduler for QT that would post an event (or signal a slot) for each scheduled action. this would leave the QT event loop untouched.

b52:
@kirkshoop Thanks for the follow up.

My current solution involves a simple timer which is scheduled on Qt's main loop and dispatches the observable events in a repetitive fashion, something like this:

```c++
rx::schedulers::run_loop runLoop;
auto worker = rx::observe_on_run_loop(runLoop);

QTimer timer;
QObject::connect(&timer, &QTimer::timeout, [&runLoop]() {
    while (runLoop.empty() && runLoop.peek().when < runLoop.now()) {
        runLoop.dispatch();
    }
});
timer.start(10);
```

Obviously this isn't ideal and has a few downsides. First of, the dispatch loop could block the main loop for too long. This is easily solvable by implementing some sort of time limit per dispatch run. Second, using a timer and thus spinning the loop every `x` ms is relatively costly for low intensity applications and increases the energy usage for no reason. Instead of a polling mechanism it would be much better to have a pushing one.

A better approach would probably involve the usage of [`QCoreApplication::postEvent()`](http://doc.qt.io/qt-5/qcoreapplication.html#postEvent) to post a dispatch event to the main loop and a custom event handler for this types of events to run the dispatch loop. However, I'm not sure how to integrate the `postEvent()` with your schedulor. Is there a way to register a handler or some sort of callback which is called once a dispatchable event is available? The following block demonstrates my idea in pseudo code:

```c++
// install the dispatch which runs the dispatch loop like above
QApplication::installEventHandler(myDispatchEventHandler);

rx::schedulers::run_loop runLoop;
auto worker = rx::observe_on_run_loop(runLoop);

// FIXME: i'm unsure if this is possible with rxcpp
worker.on_new_event([]() {
    // might make sense to check if a similar event is already
    // queued on qts main loop and to drop this event, if so
    QApplication::postEvent(spinTheDispatchLoop);
});

// starts the main loop
QApplication::exec();
```

studoot:
So, I wish I'd noticed this issue before this morning, when I went through similar thought processes, working out how I was going to integrate RxCpp and Qt. I came up with something very similar. I use two Qt signals to trigger the Rx run loop:

1. I use `QAbstractEventDispatcher::aboutToBlock` to run the dispatch loop in case there have been any newly queued schedulables during the last batch of processing Qt events. I figured this would work nicely *for me*, as most of my Rx stuff is dependent on 

2. I use a `QTimer` when there's an amount of time until the next Rx scheduled time.

    In addition, this timer is used with a 5ms interval if there is no `QAbstractEventDispatcher::instance` available when initialising.

This scheme seems to work pretty nicely - both Qt and Rx get scheduled without any noticeable delays, and CPU usage is just noise on my 3.5GHz workstation.

As @b52 alludes to, it would be useful to get notification (via callback?) when new schedulable items are queued - that could be used to trigger Rx processing rather than using `QAbstractEventDispatcher::aboutToBlock`.

I'll also mention [RxQt](https://github.com/tetsurom/rxqt), which provides some nice wrappers for creating observables from Qt signals and events...

```C++
#include <nonstd/optional.hpp> // From https://github.com/martinmoene/optional-lite/
#include <QAbstractEventDispatcher>
#include <QTimer>
#include <rxcpp/rx.hpp>

class QtRxEventsProcessor
{
public:
   QtRxEventsProcessor() : useTimerForRegularWakeup_(false)
   {
      t_.setSingleShot(true);
      t_.setTimerType(Qt::PreciseTimer);
      t_.connect(&t_, &QTimer::timeout, [&]() { onEventScheduled(); });
      if (const auto eventDispatcher = QAbstractEventDispatcher::instance())
      {
         const auto dispatcherConnection =
             QObject::connect(eventDispatcher, &QAbstractEventDispatcher::aboutToBlock, [&]() { onEventScheduled(); });
         useTimerForRegularWakeup_ = !dispatcherConnection;
      }
      else
      {
         useTimerForRegularWakeup_ = true;
      }
      onEventScheduled();
   }

private:
   void onEventScheduled()
   {
      if (const auto oTimeTillNextEvent = ProcessRxEvents())
      {
         t_.stop();
         t_.setInterval(oTimeTillNextEvent->count());
         t_.start();
      }
      else if (useTimerForRegularWakeup_)
      {
         t_.stop();
         t_.setInterval(5);
         t_.start();
      }
   }
   nonstd::optional<std::chrono::milliseconds> ProcessRxEvents() const
   {
      using namespace std::chrono;
      using namespace nonstd;
      while (!rl_.empty() && rl_.peek().when < rl_.now())
      {
         rl_.dispatch();
      }
      return rl_.empty() ? nullopt
                         : optional<std::chrono::milliseconds>(duration_cast<milliseconds>(rl_.peek().when - rl_.now()));
   }
   rxcpp::schedulers::run_loop rl_;
   QTimer t_;
   bool useTimerForRegularWakeup_;
};
```

kirkshoop:
Yes, this looks great!

I think that adding a callback to run loop that is passed the when for each call to schedule makes sense. I would merge that PR. 😉

studoot:
Hint taken! And will be followed up on...

PR #160, opened 2015-07-14
Fix gcc 4.9 error - invalid use of non-lvalue array

shivashankarp:
Not sure if I'm doing to merge and pull requests correctly.  In case I've messed up something, please let me know and I'll fix it.


kirkshoop:
Thanks for the change!

I would try to keep forks of rxcpp/master an exact replica of rxcpp/master and when making a PR start with a new branch on the fork using the latest rxcpp/master and then 'git cherry-pick ef7ce54' in that branch and make a new pull request. That way the new PR only has the new commit and no merge-commits.

kirkshoop:
I was wrong it should be `git cherry-pick c76f9ec` in this case

shivashankarp:
Thanks for the instructions, I'll do accordingly.

PR #161, opened 2015-07-15
new observer

kirkshoop:
removed dynamic_ and static_ 
Now using template specializations of observer instead.

I also added an observer specialization that stores a state and optionally takes 

Template Parameter | Function Type
-----|-----
OnNext | `void(State&, T)`
OnError | `void(State&, std::exception_ptr)`
OnCompleted | `void(State&)`

if a function is supplied it is called, if a function is not supplied then `State::on_next(T)`, `State::on_error(std::exception_ptr)`, `State::on_completed()` are called.

```cpp
rx::observable<>::create<int>(
    [&](const rx::subscriber<int>& so){
        // override the observer in the subscriber
        auto sn = rx::make_subscriber<int>(so,
            // override only on_next on the observer
            rx::make_observer<int>(so.get_observer(),
                [](rx::observer<int>& o, int v){
                    o.on_next(v + 1);
                }));
        sn.on_next(1);
        sn.on_next(2);
    });
```

due to the override this would output `[2, 3]` instead of `[1, 2]`

State does not have to be an Observer, but then all three functions must be supplied.

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-observe_on.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-observe_on.hpp
@@ -32,7 +32,6 @@ struct observe_on
     struct observe_on_observer
     {
         typedef observe_on_observer<Subscriber> this_type;
-        typedef observer_base<source_value_type> base_type;
         typedef source_value_type value_type;
         typedef rxu::decay_t<Subscriber> dest_type;
         typedef observer<value_type, this_type> observer_type;
@@ -68,7 +67,7 @@ struct observe_on
                 , destination(std::move(d))
             {
             }
-            
+
             void finish(std::unique_lock<std::mutex>& guard, typename mode::type end) const {
                 if (!guard.owns_lock()) {
                     abort();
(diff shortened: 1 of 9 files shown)


daixtrose:
Are there plan to change the name of ```as_dynamic````?

kirkshoop:
I am open to better names. What is your vote?

daixtrose:
This was only a question. You removed ```dynamic_``` and modified ```dynamic_observer```. So I wondered if the internal changes will be followed by an interface change. 

To me from user's POV ```as_dynamic()``` seems strange. I wonder if letting ```observer<T, X, Y, Z>``` inherit from ```observer<T>``` can let us get rid of the method completely, but I fear I missed some point or performance argument here. Do you try to avoid the virtual fn call? 

Can you provide the argument for not using inheritance?

kirkshoop:
The current pattern is that observer<T> is the only observer specialization with virtual functions. virtual functions block compiler inlining and other optimizations. This is a big impact to perf.

observer<T> must also allocate from the heap to store the implementation. This is another impact to perf.

rxcpp uses make_observer to construct observer specializations with no virtual methods and then these can be explicitly converted to observer<T> as appropriate.

This change was an implementation detail - the interface did not change. as_dynamic() still exists and its name intends to express that there is a run-time cost (virtual functions and an allocation are introduced).

PR #162, opened 2015-07-15
Fix gcc 4.9 erro - invalid use of non-lvalue array

shivashankarp:


kirkshoop:
merged to master d69cdd73e3e3f82d23714f782df72204cb6a611c

Thanks!

Issue #175, opened 2015-09-09
Test suite fails on ARM GCC


ksinica:
Hi, we are having some problems with RxCpp on ARM platform. RxCpp test suite is not passing, however on x86 and x86_64 platforms it runs just fine.

Target platform and toolchain details:
ARM Cortex-A8 VFPv3-D16
glibc 2.21
gcc-5.2.0
same errors were also generated on toolchain based on gcc-4.9.2 and uClibc (same hardware).

Test suite output:
http://hastebin.com/cugojuvexe.coffee

kirkshoop:
thanks for the issue and fix!

PR #176, opened 2015-09-09
Changed nextdetacher::do_unsubscribe memeber to an atomic type.

ksinica:
Changing do_unsubscribe in nextedatcher to an atomic type resolves issue #175.

### DIFF
--- a/Rx/v2/src/rxcpp/rx-subscriber.hpp
+++ b/Rx/v2/src/rxcpp/rx-subscriber.hpp
@@ -52,7 +52,7 @@ class subscriber : public subscriber_base<T>
             trace_activity().on_next_enter(*that, u);
             try {
                 that->destination.on_next(std::move(u));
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@ksinica__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, MSOTBOT;

kirkshoop:
Thanks! 

There is obviously no threading involved here. 
I suspect it is a bug with inlining. The operator and the destructor may be inlined and then do_unsubscribe gets loaded into a register for the destructor before it is written by the operator. 

Please try to change from atomic to volatile and run the tests. Hopefully that will fix the reordering issues without the perf impact of using atomic.

ksinica:
Volatile does the trick too!

Issue #189, opened 2016-01-25
operator `distinct`


gchudnov:
It looks only operator `distinct_until_changed` is implemented. But there is no `distinct`. Any reasons for that?

I can implement it (e.g. using `unordered_set`) and issue a PR if you wish..

kirkshoop:
No reason. Yes, a PR would be great!

PR #190, opened 2016-01-26
add distinct operator

gchudnov:
implemented using `std::unordered_set`.
`value` is used as a key.

### DIFF
--- a/Rx/v2/examples/doxygen/distinct.cpp
+++ b/Rx/v2/examples/doxygen/distinct.cpp
@@ -1,0 +1,15 @@
+#include "rxcpp/rx.hpp"
+
+#include "rxcpp/rx-test.hpp"
+#include "catch.hpp"
+
+SCENARIO("distinct sample"){
+    printf("//! [distinct sample]\n");
+    auto values = rxcpp::observable<>::from(1, 2, 2, 3, 3, 3, 4, 5, 5).distinct();
+    values.
+        subscribe(
+            [](int v){printf("OnNext: %d\n", v);},
+            [](){printf("OnCompleted\n");});
+    printf("//! [distinct sample]\n");
+}
+
--- a/Rx/v2/src/rxcpp/operators/rx-distinct.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-distinct.hpp
@@ -1,0 +1,81 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+#pragma once
+
+#if !defined(RXCPP_OPERATORS_RX_DISTINCT_HPP)
+#define RXCPP_OPERATORS_RX_DISTINCT_HPP
+
+#include "../rx-includes.hpp"
+
+namespace rxcpp {
+
+namespace operators {
+
+namespace detail {
+
+template<class T>
+struct distinct
+{
+    typedef rxu::decay_t<T> source_value_type;
+
+    template<class Subscriber>
+    struct distinct_observer
+    {
+        typedef distinct_observer<Subscriber> this_type;
+        typedef source_value_type value_type;
+        typedef rxu::decay_t<Subscriber> dest_type;
+        typedef observer<value_type, this_type> observer_type;
+        dest_type dest;
+        mutable std::unordered_set<source_value_type> remembered;
+
+        distinct_observer(dest_type d)
+                : dest(d)
+        {
+        }
+        void on_next(source_value_type v) const {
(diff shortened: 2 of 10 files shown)


gchudnov:
closes #189

gchudnov:
hm.. I see a problem in Clang.
Not sure how to fix it properly.

Looking at the error log, it seems when Clang tries to deduce the result type for dictinct() method, it ends up trying to instantiate something like:

```
'std::__1::hash<rxcpp::observable<int, rxcpp::dynamic_observable<int> > >'
```
for test/amb.cpp

or
```
'std::__1::hash<std::__1::vector<int, std::__1::allocator<int> > >'
```
for test/buffer.cpp

and fails.

Any ideas?

kirkshoop:
So cool to see this PR!

I took a look at the error in clang and saw
```sh
... error: implicit instantiation of undefined template 'std::__1::hash<rxcpp::observable<int, rxcpp::dynamic_observable<int> > >'
```

There is no implementation of `std::hash<>` for `rxcpp::observable<>`. This means that the `distinct()` method on observable<> must use enable_if or some other mechanism to only be 'enabled' when `std::hash<>` is defined for the `value_type` of the `observable<>`

gchudnov:
Right, but there is a problem with the following `static_assert` in [/libcxx/blob/master/include/functional](https://github.com/llvm-mirror/libcxx/blob/master/include/functional#L2557) for c++14 and above.

gchudnov:
Here is an example that triggers the assertion for c++14 and libc++:
```c++
#include <iostream>
#include <unordered_set>
#include <functional>
#include <type_traits>

class A {};
class B {};

namespace std {
  template <> struct hash<B> {
    size_t operator()(const B& x) const {
      return 0;
    }
  };
}

template<typename, typename = void>
struct is_hashable
    : std::false_type {};

template<typename T>
struct is_hashable<T, typename std::enable_if<!!sizeof(std::declval<std::hash<T>>()(std::declval<T>()))>::type>
    : std::true_type {};

int main() {
  std::cout << is_hashable<A>::value << std::endl;
  return 0;
}
```

kirkshoop:
I found one workaround, there may be others. I added this:
```cpp
namespace std {
  template <> struct hash<A>;
}
```
and the is_hashable test compiled and printed `0` when it ran.

Adding a declaration for `observable<>` at the end of rx-predef.hpp
```cpp
// disable std::hash for observable<>
namespace std {
  template <class T, class Source> struct hash<rxcpp::observable<T, Source>>;
  template <class T, class Source> struct hash<rxcpp::connectable_observable<T, Source>>;
  template <class K, class T, class Source> struct hash<rxcpp::grouped_observable<K, T, Source>>;
}
```

This appeared to work for me.

gchudnov:
Thank you!
made some changes, please review.

kirkshoop:
This rabbit hole just turned into a black hole. The workaround is not tenable since it requires every observable value type to have a std::hash declaration even if it will never use distinct(). 

This 3 year old question on [stackoverflow](http://stackoverflow.com/questions/12753997/check-if-type-is-hashable) indicates this is a longstanding issue that is still impeding usage. 

switching to a std::set might be a workaround. 
removing distinct() from observable would also work (the distinct() operator would still work).

```cpp
namespace Rx {
  using rxcpp;
  using rxcpp::sources;
  using rxcpp::operators;
}
Rx::range(0) | 
  Rx::distinct() | 
  Rx::take(5);
```

or adding a template arg to the distinct() member on observable so that it is not compiled until it is referenced (Sadly, this means that usage would get uglier `Rx.range(0).distinct<int>(); or Rx.range(0).distinct(0);`).

Thoughts?

Kirk

kirkshoop:
I tried reversing the problem and built an rxcpp whitelist of hashable types that is extensible. I pushed the result to my fork: [branch](https://github.com/kirkshoop/RxCpp/tree/distinct), [change](https://github.com/kirkshoop/RxCpp/commit/dd1965d689b9f906efb04669421c1ecbca945f7d)

This is probably the best balance of tradeoffs of the options I have seen. What do you think?

gchudnov:
You are right, that's a good trade-off.

Removing distinct() might be inconsistent with the rest of operators. And adding a template argument is ugly, I tried it yesterday. :)

kirkshoop:
ok, i made a PR for this + the whitelist to trigger appveyor and travisci. If it passes I will merge.

Thanks for your contribution!

kirkshoop:
thanks @gchudnov!

gchudnov:
Thank you!

PR #198, opened 2016-02-13
blocking_observable::blocking_subscribe

BenPope:
Improve the use of condition_variable to prevent spinning hot.


BenPope:
Looks like a timeout in docs generation :'(

kirkshoop:
Hi Ben!

Thanks for the PR!

I just got back from cpprussia conference, where I talked about Rxcpp.

These changes concern me because it took so long to find the bug and a workaround - async bugs are nasty. even if it does not repro in one configuration, it may in another. 

What tests did you run and on which platforms?

Thanks!

kirkshoop:
I would be happy to continue this discussion later, feel free to bring this back when you are ready.

PR #199, opened 2016-02-13
Reduce use of new

BenPope:
Fixes a leak in a test and use make_shared instead of shared_ptr(new ...


BenPope:
There's a couple of places where make_unique would come in handy (but it's C++14 only - perhaps a simple forwarding make_unique in rx-util would suffice).

schedulers/rx-currentthread.hpp:97,102
schedulers/rx-test.hpp:481

kirkshoop:
yes, adding make_unique in rx-util would be a good addition.

I prefer make_shared/unique everywhere, but I ran into some compiler issues in clang that I could workaround by using raw new.

I am not too concerned by this change. I'll merge it!

PR #205, opened 2016-03-09
add is_hashable check for strings

### LINKED ISSUE #204: distinct() not working with std::string
Please consider the following code:

```cpp
rxcpp::observable<std::string> obs = rxcpp::observable<>::just(std::string("Foo"));
obs.distinct(); // Error
```

The error is the `static_assert`:

> Function passed for lift() must have the signature subscriber<...>(subscriber<T, ...>)

Looking into it, `distinct()` calls `rxo::detail::distinct<std::string>` which is not enabled because `is_hashable<std::string>` seems to be false.

Now, `std::hash` [does have a specialization for `std::string`](http://en.cppreference.com/w/cpp/string/basic_string/hash), so why doesn't this work?

As a sidenote, why isn't it possible to write the following?

```cpp
rxcpp::observable<std::string> obs = rxcpp::observable<>::just("Foo");
```

gchudnov:
fixes #204

Please, review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/rx-util.hpp
+++ b/Rx/v2/src/rxcpp/rx-util.hpp
@@ -695,6 +695,18 @@ struct filtered_hash<T, typename std::enable_if<std::is_integral<T>::value>::typ
 template <class T> 
 struct filtered_hash<T, typename std::enable_if<std::is_pointer<T>::value>::type> : std::hash<T> {
 };
+template <class T>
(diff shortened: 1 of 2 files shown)


kirkshoop:
@gchudnov, thanks for the quick turnaround!
rebased, fixed build issue, added is_string<T> in #206

gchudnov:
Thank you!

PR #210, opened 2016-03-15
add debounce operator

gchudnov:
Added [debounce](http://reactivex.io/documentation/operators/debounce.html) operator.

Please, review.

Thank you,
Grigoriy


gchudnov:
Thank you!

PR #214, opened 2016-03-31
Updates for operator syntax

gchudnov:
It looks I made a couple of errors there previously, sorry.

It should be fixed now, i tested it manually.
But there are no automated tests for free functions right now. Should we add them?

Thank you,
Grigoriy


kirkshoop:
Looks good! Thanks!

kirkshoop:
@gchudnov yes, I would like to have tests for the `operator | ()` usage for all the operators. I was avoiding active use of that pattern in the beginning, but the rangev3 library has made the switch from `.operator()` to `| operator()` inevitable.

PR #216, opened 2016-04-07
remove cast to std::function inside contains operator

gchudnov:
Replaced `lambda + std::function cast` to `std::bind + std::equal_to` to avoid a virtual function call. Not sure if that is the right way.. Please reject it if it is not right.

Please review.

Thank you,
Grigoriy


gchudnov:
hm.. it looks the commit was buggy, let's close it for now :(
I'll take a look a bit later..

PR #218, opened 2016-04-12
Added timeout operator

gchudnov:
Added a variant of the [timeout](http://reactivex.io/documentation/operators/timeout.html) operator that accepts duration as a parameter.

Please review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-timeout.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-timeout.hpp
@@ -1,0 +1,218 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+#pragma once
+
+#if !defined(RXCPP_OPERATORS_RX_TIMEOUT_HPP)
+#define RXCPP_OPERATORS_RX_TIMEOUT_HPP
+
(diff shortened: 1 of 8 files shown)


Rx/v2/src/rxcpp/operators/rx-timeout.hpp
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        void on_next(T v) const {
+            auto localState = state;
+            auto work = [v, localState](const rxsc::schedulable&) {
+                auto new_id = ++localState->index;
+                auto produce_time = localState->worker.now() + localState->period;
+
+                localState->dest.on_next(v);
+                localState->worker.schedule(produce_time, produce_timeout(new_id, localState));
```
kirkshoop: This looks correct. 

FYI - if the returned subscription was stored instead of the index then this could just unsubscribe from the previous timeout and replace with the new subscription.

kirkshoop:
Looks great! Thanks!

Rx/v2/src/rxcpp/operators/rx-timeout.hpp
gchudnov: Got it, thank you!

PR #227, opened 2016-06-01
remove some lambdas

kirkshoop:
remove some lambdas that were contributing to longer compile times.

### DIFF
--- a/Rx/v2/src/rxcpp/rx-observable.hpp
+++ b/Rx/v2/src/rxcpp/rx-observable.hpp
@@ -466,6 +466,36 @@ public:
     }
 };
 
+namespace detail {
+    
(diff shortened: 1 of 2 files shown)


kirkshoop:
clang apt server is disabled due to overuse.. ignoring those failures..

gchudnov:
just wondering, what was the impact?

Thank you,
Grigoriy

kirkshoop:
I used some instructions sent to me in email to install templight. This records template instantiations during compilation.

I was hoping that I could find some changes that would speed up compilation of rxcpp. 

What I found was that a huge number of instantiations were caused by two lambdas. I made dedicated functors for each. 

I do not see the big improvement in compile time that I hoped for, so I either need to look at the templight data differently or try another tool.

PR #244, opened 2016-08-31
add sequence_equal operator

gchudnov:
Added [sequence_equal](http://reactivex.io/documentation/operators/sequenceequal.html) operator.

In this commit, I used `!=` to check items for inequality.
Going to add a method with a binary predicate in one of the next commits.

Please, review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-sequence_equal.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-sequence_equal.hpp
@@ -1,0 +1,215 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+#pragma once
+
+#if !defined(RXCPP_OPERATORS_RX_SEQUENCE_EQUAL_HPP)
+#define RXCPP_OPERATORS_RX_SEQUENCE_EQUAL_HPP
+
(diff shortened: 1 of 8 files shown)


Rx/v2/src/rxcpp/operators/rx-sequence_equal.hpp
```diff
@@ -0,0 +1,215 @@
...
+
+        auto check_equal = [state]() {
+            if(!state->source_values.empty() && !state->other_values.empty()) {
+                auto x = std::move(state->source_values.front());
+                state->source_values.pop_front();
+
+                auto y = std::move(state->other_values.front());
+                state->other_values.pop_front();
+
+                if (x != y) {
```
kirkshoop: when you change to the binary predicate please change this to !pred(x, y) or if this line of code remains, change to !(x == y)

In either case I would like == to be the default operator used instead of !=

kirkshoop:
Looks great - Thanks!

Rx/v2/src/rxcpp/operators/rx-sequence_equal.hpp
gchudnov: Sure, I'll do it

gchudnov:
Thank you!

PR #253, opened 2016-09-23
Removed naming conflict with Qt's emit define

marcobusemann:
The alternative would be to configure Qt to not define emit. But obviously we are not able to do that in every project.


msotclas:
Hi __@marcobusemann__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, MSOTBOT;

kirkshoop:
Thanks!

PR #254, opened 2016-09-27
decouple zip operator from observable

kirkshoop:
made changes to the operator implementation pattern. using zip as the
first example.

- allows rx-lite.hpp to not include the operators and instead have the app explicitly include only the operators needed.
- allows all zip related logic to be only in the rx-zip.hpp. observable only has a forwarding function named zip and operators define zip_tag and its include_header member.
- improves docs by linking the observable and free operator methods to the single doc at file scope.
- operator_member_missing overloads allow static asserts to help usage.
- the include_header member of the zip_tag allows a static assert to ask for the zip file to be included when omitted in rx-lite mode

@gchudnov, as a regular operator contributor, I would especially appreciate your feedback here. Is this an improvement? Would you do anything differently?

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-zip.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-zip.hpp
@@ -7,63 +7,130 @@
 
 #include "../rx-includes.hpp"
 
+/*! \file rx-zip.hpp
+
+    \brief Bring by one item from all given observables and select a value to emit from the new observable that is returned.
+
+    \tparam AN  types of scheduler (optional), aggregate function (optional), and source observables
+
+    \param  an  scheduler (optional), aggregation function (optional), and source observables
+
+    \return  Observable that emits the result of combining the items emitted and brought by one from each of the source observables.
+
+    If scheduler is omitted, identity_current_thread is used.
+
+    If aggregation function is omitted, the resulting observable returns tuples of emitted items.
+
+    \sample
+
+    Neither scheduler nor aggregation function are present:
+    \snippet zip.cpp zip sample
+    \snippet output.txt zip sample
+
+    Only scheduler is present:
+    \snippet zip.cpp Coordination zip sample
+    \snippet output.txt Coordination zip sample
+
+    Only aggregation function is present:
+    \snippet zip.cpp Selector zip sample
+    \snippet output.txt Selector zip sample
+
+    Both scheduler and aggregation function are present:
+    \snippet zip.cpp Coordination+Selector zip sample
+    \snippet output.txt Coordination+Selector zip sample
+*/
+
 namespace rxcpp {
 
 namespace operators {
 
 namespace detail {
 
-template<class T>
+template<class Observable>
 struct zip_source_state
 {
+    using value_type = rxu::value_type_t<Observable>;
     zip_source_state() 
         : completed(false) 
     {
     }
-    std::list<T> values;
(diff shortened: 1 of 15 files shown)


gchudnov:
Looks great!
I like the way the related operator logic and docs are localized to a specific file. It should simplify the development and the usage of the library.
And better diagnostic messages on errors definitely helps a lot.

I'll give more feedback, just need some more time to play with it..

Rx/v2/src/rxcpp/operators/rx-zip.hpp
```diff
@@ -200,55 +241,120 @@ struct zip : public operator_base<rxu::value_type_t<zip_traits<Coordination, Sel
     }
```
gchudnov: Trying to implement an operator in a new way..

```struct zip : public operator_base```
^^
Is that the preferred way to create an operator?

**Comment by gchudnov** [comment]
Just implemented [debounce](https://github.com/gchudnov/RxCpp/commit/5330a34c6503822199b432ca6ec5fa45205af07b) in a new way (draft, it looks I messed up coordination a little and need to improve the docs).

The implementation is a bit complex, however with the help of better diagnostic messages it *might* be easier to use the library.

The only thing that worries me is the lack of explicit function signatures. It might be a bit complex to use the library for a user.
The signature of an operator is buried inside:
```
auto debounce(AN&&... an) const
```
and the overloads of:
```
Result observable_member_for(....
```

However, the better documentation might fix that.

File: Rx/v2/src/rxcpp/operators/rx-zip.hpp (hunk 1 above)
Comment (kirkshoop) [inline Rx/v2/src/rxcpp/operators/rx-zip.hpp]: it has been :)

I haven't yet heard of an issue where having the members be public induced someone to muck with them and break the operators contract. That would push me over the edge to making the members private.

I do however consider the members to be implementation details that could change in minor releases. This means that the members should be private now.

Conclusion - I am not going to require the members to be private, but I now recommend that they are made private.

**Comment by kirkshoop** [comment]
debounce looks nice!

I have a problem on VS2013 that is preventing me from merging zip. I hope to workaround that soon.

Yes, the generic function signatures require more documentation. The `observable_member_missing` overloads can also be used explain what is expected at compile-time. 

The generic overloads also guarantee that both the `.zip` and `| rxo::zip` paths have the exact same overloads.

File: Rx/v2/src/rxcpp/operators/rx-zip.hpp [hunk 2]
```diff
@@ -7,63 +7,130 @@
...
         return val;
     }
 };
 
-template<class Coordination, class Selector, class... ObservableN>
-struct zip_traits {
-    typedef std::tuple<ObservableN...> tuple_source_type;
-    typedef std::tuple<zip_source_state<typename ObservableN::value_type>...> tuple_source_values_type;
+template<class... AN>
+struct zip_invalid_arguments {};
```
gchudnov: @kirkshoop  

In a [PR](https://github.com/Reactive-Extensions/RxCpp/pull/258) I made a copy of `zip_invalid_arguments`, `zip_invalid`, `zip_invalid_t`.

And it looks these structures might be similar for all operators. What do you think about extracting them?

kirkshoop: I decided not to since some compilers did not trigger the static assert and this makes the operator and arguments clear in the debug output when the overload is missing

PR #258, opened 2016-10-14
decouple debounce from observable

gchudnov:
Updated to follow the new operator implementation pattern.
Please review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-debounce.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-debounce.hpp
@@ -7,209 +7,299 @@
 
 #include "../rx-includes.hpp"
 
+/*! \file rx-debounce.hpp
+
+    \brief  Return an observable that emits an item if a particular timespan has passed without emitting another item from the source observable.
+
+    \tparam Duration      the type of time interval
+    \tparam Coordination  the type of the scheduler
+
+    \param period        the period of time to suppress any emitted items
+    \param coordination  the scheduler to manage timeout for each event
+
+    \return  Observable that emits an item if a particular timespan has passed without emitting another item from the source observable.
+
+    \sample
+    \snippet debounce.cpp debounce sample
+    \snippet output.txt debounce sample
+*/
+
 namespace rxcpp {
 
 namespace operators {
 
 namespace detail {
 
-template<class T, class Duration, class Coordination>
-struct debounce
-{
-    typedef rxu::decay_t<T> source_value_type;
-    typedef rxu::decay_t<Coordination> coordination_type;
-    typedef typename coordination_type::coordinator_type coordinator_type;
-    typedef rxu::decay_t<Duration> duration_type;
+template<class... AN>
+struct debounce_invalid_arguments {};
 
-    struct debounce_values
+template<class... AN>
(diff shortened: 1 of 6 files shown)


Rx/v2/src/rxcpp/operators/rx-debounce.hpp
```diff
@@ -7,209 +7,299 @@
...
-        return      debounce_observer<Subscriber>::make(std::move(dest), initial);
+        auto coordinator = initial.coordination.create_coordinator(scbr.get_subscription());
+        auto state = std::make_shared<debounce_state_type>(initial, std::move(coordinator), std::move(scbr));
+        subscribe_one(state);
     }
 };
 
-template<class Duration, class Coordination>
+template<class... AN>
 class debounce_factory
```
kirkshoop: You can use operator_factory instead.

Rx/v2/test/operators/debounce.cpp
```diff
@@ -29,7 +31,7 @@ SCENARIO("debounce - never", "[debounce][operators]"){
 
             THEN("there was 1 subscription/unsubscription to the source"){
                 auto required = rxu::to_vector({
-                    on.subscribe(200, 1001)
+                    on.subscribe(200, 1000)
```
kirkshoop: The test changes worry me a bit. Looks like some things are getting scheduled that were not before

Rx/v2/src/rxcpp/operators/rx-debounce.hpp
```diff
@@ -7,209 +7,299 @@
...
+    typedef rxu::decay_t<Duration> duration_type;
     typedef rxu::decay_t<Coordination> coordination_type;
+    typedef rxu::decay_t<Observable> source_type;
+    typedef typename source_type::value_type source_value_type;
+    typedef source_value_type value_type;
+};
+
+template<class Coordination, class Duration, class Observable>
+struct debounce : public operator_base<rxu::value_type_t<debounce_traits<Coordination, Duration, Observable>>>
+{
```
kirkshoop: This changed the operator from a lifter to an observable. I need to figure out what the pattern looks like for lifters because they are preferable. 

If we can figure out how to change the scheduling to match the previous test behavior then this change is good.

Thank you for helping me validate the pattern!

Rx/v2/test/operators/debounce.cpp
gchudnov: right, when changing the operator from a lifter to an observable, I made some changes :(

To be honest, it looks I'm still confused about some things and not sure what is the proper way to implement them.

For example, in the [following code at master](https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/operators/rx-debounce.hpp#L90): 
the unsubscribe is scheduled and gets + 1 added to its time (get 1001 instead of 1000). However, is it ok to unsubscribe right away without scheduling it?

Rx/v2/src/rxcpp/operators/rx-debounce.hpp
kirkshoop: I am trying out a pattern for the lifters in #260 
Hopefully VS will compile :)

gchudnov:
Re-implemented using a pattern for lifters in a new PR, closing this one.

PR #262, opened 2016-10-22
decouple debounce from observable

gchudnov:
Reimplemented using the pattern for lifters.
Please review.

Thank you,
Grigoriy


kirkshoop:
Looks great! Thanks for your help!

PR #264, opened 2016-10-28
Decouple filter from observable

gchudnov:
Decoupled `filter` from `observable` without implementation changes.
Please, review.

btw,
It might be better to extract [these](https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/test/operators/filter.cpp#L46-L66) conditional testing into separate tests without `#if`. In this way both cases will be verified. I can make a separate PR if you wish.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-filter.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-filter.hpp
@@ -95,20 +95,6 @@ struct filter
     }
 };
 
-template<class Predicate>
-class filter_factory
-{
-    typedef rxu::decay_t<Predicate> test_type;
-    test_type predicate;
-public:
-    filter_factory(test_type p) : predicate(std::move(p)) {}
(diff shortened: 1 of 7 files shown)


Rx/v2/src/rxcpp/operators/rx-filter.hpp
```diff
@@ -86,14 +111,35 @@ class filter_factory
 
 }
 
-template<class Predicate>
-auto filter(Predicate&& p)
-    ->      detail::filter_factory<Predicate> {
-    return  detail::filter_factory<Predicate>(std::forward<Predicate>(p));
```
kirkshoop: I think that `detail::filter_factory` can be deleted.

kirkshoop:
@gchudnov yes, separating at least one test that uses operator| would be better than the #ifdef.

PR #267, opened 2016-10-30
decouple distinct from observable

gchudnov:
Decouple distinct from observable.
No implementation changes.

Please, review

Thank you,
Grigoriy


kirkshoop:
Nice! thanks for the work on MSVC 👍

PR #275, opened 2016-11-23
decouple all from observable

gchudnov:
Decoupled [all](http://reactivex.io/documentation/operators/all.html) from observable
Please review

Thank you,
Grigoriy


kirkshoop:
Looks Great!

PR #276, opened 2016-11-24
decouple exists and contains from observable

gchudnov:
Decoupled [exists and contains](http://reactivex.io/documentation/operators/contains.html) from observable.
Please, review

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-any.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-any.hpp
@@ -2,6 +2,30 @@
 
 #pragma once
 
+/*! \file rx-any.hpp
+
+    \brief Returns an Observable that emits true if any item emitted by the source Observable satisfies a specified condition, otherwise false. Emits false if the source Observable terminates without emitting any item.
+
+    \tparam Predicate the type of the test function.
+
+    \param p the test function to test items emitted by the source Observable.
+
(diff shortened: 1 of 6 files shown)


Rx/v2/src/rxcpp/operators/rx-any.hpp
```diff
@@ -78,37 +113,99 @@ struct any
...
+template<class... AN>
+auto contains(AN&&... an)
+->     operator_factory<contains_tag, AN...> {
+    return operator_factory<contains_tag, AN...>(std::make_tuple(std::forward<AN>(an)...));
 }
+
 }
 
+template<>
+struct member_overload<exists_tag>
```
kirkshoop: I do not see any specialization for `any_tag`, did I miss it?
Do we have any tests for `any()` or only `exists()` and `contains()`?

kirkshoop: Looks Great!

Perhaps we can just change one of the `exists()` test to use `any()` to ensure that it does not get broken later.

kirkshoop:
Thank You!

Rx/v2/src/rxcpp/operators/rx-any.hpp
gchudnov: Sure, I'll do in in one of the next PRs

PR #280, opened 2016-11-30
decouple finally from observable

gchudnov:
Decoupled finally from observable, added tests.

Please review.

Thank you,
Grigoriy


kirkshoop:
Looks great!
Shouldn't finally() have been the last to be converted? Or maybe last() should have been the last to be converted.. 😃

gchudnov:
True, haven't thought about it... :laughing:

PR #281, opened 2016-12-06
decouple map from observable

gchudnov:
Decopuled [map](http://reactivex.io/documentation/operators/map.html) from obervable, added tests.

Please, review

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-map.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-map.hpp
@@ -2,6 +2,21 @@
 
 #pragma once
 
+/*! \file rx-map.hpp
+
+    \brief For each item from this observable use Selector to produce an item to emit from the new observable that is returned.
+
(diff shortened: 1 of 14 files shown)


gchudnov:
@kirkshoop,
I noticed that [first_tag](https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/rx-operators.hpp#L243) doesn't inherit from `reduce_tag`.

Guess it should be:
```
struct first_tag : reduce_tag {};
```

Is that intentional?

kirkshoop:
@gchudnov - I think you are right. we do not depend on that for the observable_member specializations, but we do for the "please include" static_assert

Thanks!

kirkshoop:
nice! this was a big one.

Now that it is so easy, I want to add `transform()` as an alias of `map()`. I want to make it obvious which std algo is the counterpart of which Rx algo. I should have added `accumulate()` when I did `reduce()`

Issue #284, opened 2016-12-08
distinct_until_changed operator for any class


Ravirael:
Static assertion failes if someone tries to use `distinct_until_changed` with something different than primitive or string.

```
#include <rxcpp/rx.hpp>

using namespace rxcpp;
using namespace rxcpp::sources;
using namespace rxcpp::operators;
using namespace rxcpp::util;

struct A {
    int i;

    bool operator!=(const A& a) const {
        return i != a.i;
    }

    bool operator==(const A& a) const {
        return i == a.i;
    }
};

int main() {
    auto o = from(A{0}, A{1}, A{1}, A{2}).distinct_until_changed();
}
```

The example above does not compile. Even defining `std::hash` specialization for `A` doesn't help. That's probably an issue with `rxcpp::util::is_hashable` and `rxcpp::util::filtered_hash`. Why does `distinct_until_changed` require `source_value_type` to be hashable anyway?

kirkshoop:
Thank you for the report!

`distinct_until_changed()` does not depend on the type being hashable. only `distinct()` requires this and that is due to it using `unordered_set<>`.

I will need to see the compiler output to find the actual cause.

Ravirael:
@kirkshoop The output from GCC 6.2.1 compiled with `-std=c++14`:

```
In file included from /usr/include/rxcpp/rx-includes.hpp:189:0,
                 from /usr/include/rxcpp/rx.hpp:8,
                 from /home/rames/Projects/Clion/RxPlayground/main.cpp:1:
[...]

PR #285, opened 2016-12-08
fixes issue 284

gchudnov:
A fix for #284 
distinct_until_changed:
* Removed the invalid requirement for a type to be hashable.
* `operator==` is used to determine if two elements are equal.
* added test


kirkshoop:
This change is good!

The osx build is stuck in the queue on travisci, but I am going to pull the trigger based on the rest of the completed builds.

gchudnov:
Thank you!

PR #286, opened 2016-12-09
add support for C++ coroutines

kirkshoop:


Rx/v2/src/rxcpp/rx-coroutine.hpp
```diff
@@ -0,0 +1,194 @@
...
+};
+
+template<typename Source>
+struct co_observable_inc_awaiter
+{
+    bool await_ready() {
+        return false;
+    }
+
+    bool await_suspend(const coroutine_handle<>& handle) {
```
GorNishanov: I would drop const& so that the parameter type is just coroutine_handle<> . The coroutine_handle is just a fancy template around a void* with no destructor, so passing it by value is cheaper.

Rx/v2/src/rxcpp/rx-coroutine.hpp
```diff
@@ -0,0 +1,194 @@
...
+    using value_type=typename iterator::value_type;
+
+    explicit co_observable_iterator_awaiter(const Source& o) : it(o) {
+    }
+
+    bool await_ready() {
+        return false;
+    }
+
+    void await_suspend(const coroutine_handle<>& handle) {
```
GorNishanov: Drop const&

PR #289, opened 2016-12-14
decouple on_error_resume_next from observable

gchudnov:
Decoupled `on_error_resume_next` from observable.
In a separate commit made `first_tag` to inherit from `reduce_tag`.

Please, review.

Thank you,
Grigoriy


kirkshoop:
Thank you! 👍

PR #295, opened 2016-12-17
decouple sequence_equal from observable

gchudnov:
Decoupled `sequence_equal` from observable.
Please review.

FYI:
[Here](https://github.com/Reactive-Extensions/RxCpp/compare/master...gchudnov:sequence_equal-o2?expand=1#diff-5291158f99c1fafd323041704e108a21R792) I ported [std::negation](http://en.cppreference.com/w/cpp/types/negation) from C++17 and used it [for overload selection](https://github.com/Reactive-Extensions/RxCpp/compare/master...gchudnov:sequence_equal-o2?expand=1#diff-c3362080e2d5eeea4f8eee60638ba9ebR236). If that is not OK and should be done in a different way, let me know.

Thank you,
Grigoriy.


kirkshoop:
Negation is perfect, thanks!

PR #301, opened 2016-12-26
decouple timestamp from observable

gchudnov:
Decoupled `timestamp` from observable, please review.

and Merry Christmas! :christmas_tree: 

Thank you,
Grigoriy


kirkshoop:
Merry Christmas!🎄

PR #311, opened 2016-12-31
decouple window from observable

gchudnov:
Decouple `window` from observable,
please review.

Thank you,
Grigoriy


kirkshoop:
thank you!

PR #317, opened 2017-01-06
decouple take_last from observable

gchudnov:
Another minor update..
Decouple `take_last` from observable, please review.

Thank you,
Grigoriy


kirkshoop:
Nothing minor about your contributions! We are so close to having all the operators decoupled, thanks to you!

PR #320, opened 2017-01-09
decouple skip_until from observable

gchudnov:
Decoupled `skip_until` from observable.
To make it similar to `take_until`, added a version of skip_until with a time point.
Please review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-skip_until.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-skip_until.hpp
@@ -2,6 +2,28 @@
 
 #pragma once
 
+/*! \file rx-skip_until.hpp
+
+    \brief Make new observable with items skipped until on_next occurs on the trigger observable or until the specified time.
+           skip_until takes (TriggerObservable, optional Coordination) or (TimePoint, optional Coordination)
+
+    \tparam  TriggerSource  the type of the trigger observable.
+    \tparam  Coordination   the type of the scheduler (optional).
+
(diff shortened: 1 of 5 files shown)


gchudnov:
@kirkshoop 
btw, do you mind if a make std::chrono::milliseconds in the [test scheduler](https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/schedulers/rx-test.hpp#L48-L53) a typedef, something like:

```
typedef std::chrono::milliseconds duration_type;
```

In this way tests can be written in terms of that `duration_type`, instead of specifying milliseconds [explicitly](https://github.com/Reactive-Extensions/RxCpp/pull/320/files#diff-b7592a63bb4a4c2dfc8252e0e5e6f2c8R580):

Instead of:
```
 clock_type::time_point t(std::chrono::milliseconds(225));
```

there will be:
```
 clock_type::time_point t(duration_type(225));
```

kirkshoop:
@gchudnov, yes, I think removing the explicit milliseconds from the tests is a good idea. 

Perhaps, instead of a typedef, we could expose to_time_point() on the test scheduler by calling the same method on the shared state. 

Thus auto t = sc.to_time_point(225); would replace the manual construction of the time_point from the duration.

kirkshoop:
Thank you for making the _until methods consistent as well!

gchudnov:
done #321, please take a look

PR #321, opened 2017-01-09
exposed to_time_point on the test scheduler

gchudnov:
Exposed `to_time_point` on the test scheduler, updated `skip_until` and `take_until` tests,
please review.

Thank you,
Grigoriy


kirkshoop:
That looks nicer to me, what do you think?

Thank you!

gchudnov:
agreed, that's definitely better and allows to write tests without relying on implementation details.

PR #323, opened 2017-01-11
decouple scan from observable

gchudnov:
Decoupled `scan` from observable,
please review.

Thank you,
Grigoriy


kirkshoop:
Thank you!

Issue #324, opened 2017-01-12
Issue with lifetime of subject's observers


wutiantong:
The core problem is : 
**A member named `current_completer` would still retain unsubscribed observers of subject, until the subject's subscriber meet the next `on_next()`, or the end of lifetime.**

In consequence : 
1. If `connectable_observable`(aka `publish()`) was disconnected, `unsubscribe()` can't release it's observers.
2. When using `publish().ref_count()`, the last-unsubscribed observer would never be released.

I have made a patch for this problem, if need, I would like to send a PR.

gchudnov:
Thank you for the report!

To make sure we're on the same page, could you provide some code to reproduce the problem?
And yes, PRs are always welcomed!

wutiantong:
```C++
#include "test.h"

struct Observer {
    std::string tag;
    Observer(std::string t): tag(t) {}
    ~Observer() { std::cout << "~Observer: " << tag << '\n'; }
};

rx::observer<long> make_observer(std::string tag) {
    auto o = std::make_shared<Observer>(tag);
    return rx::make_observer_dynamic<long>([o](long){});
}

SCENARIO("connectable_observable", "[issue324]") {
    GIVEN("a interval connectable_observable") {
        using namespace std::chrono;
        auto sc = rxsc::make_new_thread();
        auto so = rx::synchronize_in_one_worker(sc);
        auto start = sc.now() + seconds(1);
        auto period = seconds(1);
        auto connectable = rx::observable<>::interval(start, period, so).publish();
        rx::composite_subscription connection;
        connectable.connect(connection);
        WHEN("make 3 observers") {
            auto cs1 = connectable.subscribe(make_observer("ob1"));
            auto cs2 = connectable.subscribe(make_observer("ob2"));
            auto cs3 = connectable.subscribe(make_observer("ob3"));

            THEN("keep connection, unsubscribe observers one by one") {
                std::cout << "keep connection, unsubscribe observers one by one\n";

                std::this_thread::sleep_for(seconds(2));
                std::cout << "unsubscribe ob1 after 2s\n";
                cs1.unsubscribe();

                std::this_thread::sleep_for(seconds(2));
                std::cout << "unsubscribe ob2 after 2s\n";
                cs2.unsubscribe();

                std::this_thread::sleep_for(seconds(2));
                std::cout << "unsubscribe ob3 after 2s\n";
                cs3.unsubscribe();

                std::this_thread::sleep_for(seconds(1));
[...]

PR #325, opened 2017-01-12
Patch for issue with lifetime of subject's observers #324

wutiantong:
Notice: `lock()`method of `std::weak_ptr` is [executed atomically](http://en.cppreference.com/w/cpp/memory/weak_ptr/lock)


kirkshoop:
@wutiantong 

this passed the concurrency tests, looks great!

I would like to have tests for this leak, your comment in #324 is a great start. Would you like to add them in a separate PR?

PR #328, opened 2017-01-13
decouple tap from observable

gchudnov:
Decoupled `tap` from observable,
please review

Thank you,
Grigoriy


kirkshoop:
Thank you!

PR #329, opened 2017-01-14
decouple switch_if_empty, default_if_empty from observable

gchudnov:
Decoupled `switch_if_empty`, `default_if_empty` from observable,
please review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-switch_if_empty.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-switch_if_empty.hpp
@@ -2,6 +2,21 @@
 
 #pragma once
 
+/*! \file rx-switch_if_empty.hpp
+
+    \brief If the source Observable terminates without emitting any items, emits items from a backup Observable.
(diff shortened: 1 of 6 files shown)


gchudnov:
@kirkshoop,
MSVC2013 fails with `fatal error C1045: compiler limit : linkage specifications nested too deeply`
Have you bumped into it?

Don't have MSVC2013 at hand, not sure what is the best way to fix it... :slightly_frowning_face: 
Probably by replacing `from` to `iterate`?

kirkshoop:
@gchudnov - yes this is a bad fixed compiler limit. There are only a fixed number of nested decltype() allowed. I was eventually able to fix this recently by finding a way to remove a level of decltype(). I had to find the offending decltype by laboriously reducing the code until I found which one was too deeply nested.

gchudnov:
should work now... :smile:

kirkshoop:
Nice! Thank you!

PR #330, opened 2017-01-17
decouple switch_on_next from observable

gchudnov:
Decoupled `switch_on_next` from observable, please review.

Thank you,
Grigoriy


kirkshoop:
Thank you! start_with, the multicast set, subscribe/observe_on and flat/concat_map are all that appear to be pending - Amazing!

gchudnov:
yea, we're close...

PR #332, opened 2017-01-19
decouple flat_map from observable

gchudnov:
Decouple `flat_map` from observable, added some tests.
Please review.

Thank you,
Grigoriy


kirkshoop:
Maps complete!  Thank you!

PR #333, opened 2017-01-20
decouple start_with from observable

gchudnov:
Decoupled `start_with` from observable, added tests.
Please review...

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/examples/linesfrombytes/main.cpp
+++ b/Rx/v2/examples/linesfrombytes/main.cpp
@@ -13,6 +13,7 @@
 #include <rxcpp/operators/rx-ref_count.hpp>
 #include <rxcpp/operators/rx-window.hpp>
 #include <rxcpp/operators/rx-window_toggle.hpp>
+#include <rxcpp/operators/rx-start_with.hpp>
 namespace Rx {
 using namespace rxcpp;
 using namespace rxcpp::sources;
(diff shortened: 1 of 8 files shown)


gchudnov:
It seems I bumped into the same problem with VC2013 again:
`fatal error C1045: compiler limit : linkage specifications nested too deeply`
Will need some time to figure it out...

gchudnov:
slightly changed the implementation, hope it fixes msvc2013.
The error `fatal error C1045: compiler limit : linkage specifications nested too deeply` is quite painful.. :disappointed:

kirkshoop:
Yes, I know. When I rewrite rxcpp I will use at least c++14 and vs2013 will not be supported!

kirkshoop:
Thank you!

Perhaps I should poll on twitter for vs2013 usage?

gchudnov:
> Perhaps I should poll on twitter for vs2013 usage?

Guess, there is no need if the next major version won't use it anyway...
There might be some issues with vs2013 in v2, but I don't think they will be the show-stoppers..

I've seen the v3 in your repo. What stops making it the next major version? Or is that thing is a pure experiment?

kirkshoop:
Some parts of v3 need another pass (context and scheduler) the rest I am pretty happy with. 

I think that the observable decoupling will be the last big transform in the v2 code. Hopefully v3 can be polished up and put into the official repo as the eventual replacement during this year.

PR #337, opened 2017-01-25
Add callback for when a task is added to a run-loop schedule

studoot:
See #154 for motivation for this PR...

Implementation decisions:

1. I've added a single callback as a `std::function<void()>`:
    1. I don't think the task details are too relevant for the callback
    2. I think there's only likely to be one callback needed to be registered at any one time, given the use case of integrating the Rx loop with external event loops.
2. The callback can be disconnected by setting it to `std::function<void()>{}`

### DIFF
--- a/Rx/v2/src/rxcpp/schedulers/rx-runloop.hpp
+++ b/Rx/v2/src/rxcpp/schedulers/rx-runloop.hpp
@@ -35,6 +35,7 @@ struct run_loop_state : public std::enable_shared_from_this<run_loop_state>
     mutable std::mutex lock;
     mutable queue_item_time q;
     recursion r;
+    std::function<void()> add_task_callback;
 };
 
 }
@@ -79,6 +80,9 @@ private:
                 std::unique_lock<std::mutex> guard(st->lock);
                 st->q.push(detail::run_loop_state::item_type(when, scbl));
                 st->r.reset(false);
+                std::function<void()> callback{st->add_task_callback};
+                guard.unlock(); // So we can't get attempt to recursively lock the state
+                if (callback) callback();
             }
         }
     };
@@ -188,6 +192,11 @@ public:
 
     scheduler get_scheduler() const {
         return make_scheduler(sc);
+    }
+
+    void set_add_task_callback(std::function<void()> const& f) {
+        std::unique_lock<std::mutex> guard(state->lock);
+        state->add_task_callback = f;
     }
 };
 
--- a/Rx/v2/src/rxcpp/schedulers/rx-runloop.hpp
+++ b/Rx/v2/src/rxcpp/schedulers/rx-runloop.hpp
@@ -35,7 +35,7 @@ struct run_loop_state : public std::enable_shared_from_this<run_loop_state>
     mutable std::mutex lock;
     mutable queue_item_time q;
     recursion r;
-    std::function<void()> add_task_callback;
+    std::function<void(clock_type::time_point)> notify_earlier_wakeup;
 };
 
 }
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@studoot__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, MSOTBOT;

Rx/v2/src/rxcpp/schedulers/rx-runloop.hpp
```diff
@@ -79,6 +80,9 @@ struct run_loop_scheduler : public scheduler_interface
                 std::unique_lock<std::mutex> guard(st->lock);
                 st->q.push(detail::run_loop_state::item_type(when, scbl));
                 st->r.reset(false);
+                std::function<void()> callback{st->add_task_callback};
+                guard.unlock(); // So we can't get attempt to recursively lock the state
+                if (callback) callback();
```
kirkshoop: `schedule()` is called from other threads in order to schedule work on the thread that owns the `runloop`. 

`callback()` will be called from other threads. It is not okay to dispatch any events from the callback since they would be running on the wrong thread. The only thing that the callback can do is reset the timer to an earlier value if the new item is supposed to occur earlier than the previous timer was set. This would be easier if the callback was passed the new time.

If so, perhaps it would be better if the callback was only called when the new `when` is earlier than the previous queue top. In that case, the name of the callback might be `notify_next_wakeup` instead of `add_task_callback`.

kirkshoop: `schedule()` is called from threads other than the thread the owns the `runloop`.

`callback()` is also called from other threads. The callback must not dispatch `runloop` items. The only thing the callback could safely do is set a timer to trigger at an earlier time than it currently is set.

passing the new `when` would make that easier.
only calling the callback if the new time is earlier than the previous top item would be reasonable as well.

If those changes are made then the name of the callback might change as well. Perhaps `notify_when_changed`?

kirkshoop:
:) I was sure that github had lost my comment - it just disappeared. So I wrote it again!

studoot:
Hmmmm - without an easy way of signalling the run loop's thread, I think this needs a bit more thought.

One thing I was thinking of was to just provide more information about the state of the queue (are there tasks ready for dispatch or newly added since the last dispatch?). Or to document that the callback is *purely* to wake the run loops thread (I know how to do *that* with Qt...)... How does that sound?

kirkshoop:
> One thing I was thinking of was to just provide more information about the state of the queue (are there tasks ready for dispatch ...

I think that passing the `when` to the callback and only calling the callback if the item replaced the front/top of the queue will indicate that a task is ready for dispatch and what time the QT timer should be created to process the item on the runloop thread. 

I think that a small change to what you have in this PR will create this semantic.

> ...  or newly added since the last dispatch?)). Or to document that the callback is purely to wake the run loops thread (I know how to do that with Qt...)... How does that sound?

IMO, 'newly added' should not signal the runloops thread. Many times the new item will be in the future, or at the same time as the item already at the front/top of the runloop. Waking the thread to run nothing or waking the thread repeatedly for the same time, when each item is scheduled, would be very chatty.

Do you think that you could do what you need in QT with the changes to the callback I proposed?

studoot:
To answer your question - yes, and here's an updated PR... This works with Qt and (if anything) simplifies that code.

Rationale for some of the change:

1. I've made the `run_loop`'s `clock_type` public, to allow its use in the definition of the callback function's type.
2. The callback is triggered while the mutex is locked. This is so that the scheduler queue can't be manipulated within the callback, encouraging minimal code in the callback.

kirkshoop:
I am happy that this works well for QT, that gives me confidence.

Thank you for this change!

PR #338, opened 2017-01-25
decouple publish, publish_synchronized, multicast from observable

gchudnov:
Decoupled `publish, publish_synchronized, multicast` from observable,
please review.

`publish_synchronized` was a bit tricky since [rx-synchronize](https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/subjects/rx-synchronize.hpp#L219) references it.
I added #include "publish.hpp" directives to some operators (e.g. [here](https://github.com/Reactive-Extensions/RxCpp/compare/master...gchudnov:publish-o2?expand=1#diff-4c3f459d8f4ccd498ab815ca3c67eeb6R43)), otherwise it didn't compiled.

Please, take a look if it is ok.

Thank you,
Grigoriy


Rx/v2/src/rxcpp/rx-operators.hpp
```diff
@@ -97,8 +97,6 @@ class operator_factory
 
 #include "operators/rx-connect_forever.hpp"
 #include "operators/rx-lift.hpp"
-#include "operators/rx-multicast.hpp"
-#include "operators/rx-publish.hpp"
```
kirkshoop: I would like to think about these dependencies on publish/multicast longer.

Until then, please try the build with the original include structure (add these two includes back and remove all the includes of multicast and publish that were added).

Thank you!

kirkshoop:
@gchudnov I looked yesterday and decided to sleep on it.. I am still thinking. :)
I think we can just keep them in the always-included set for now - like subscribe, lift etc..

Rx/v2/src/rxcpp/rx-operators.hpp
gchudnov: If added back, there will be a compile error:
```
rx-multicast.hpp:93:30: error: ‘multicast_tag’ was not declared in this scope
```
since the multicast_tag is used inside of the included file, but defined below these lines..
I can try to put them to the end of the file after all of the `struct YYY_tag {` definitions.

kirkshoop: Yes of course. 
Yes try putting them after the tag is defined.

gchudnov: The include structure should be restored, please take a look at the last commit.

kirkshoop:
Thank you!

I think that including multicast and publish in includes.hpp is not needed, but it is not a problem.

gchudnov:
> ...including multicast and publish in includes.hpp is not needed..

My fault, completely forgot about it, I'll remove in a separate PR.
Thank you for merging!

PR #340, opened 2017-01-27
decouple connect_forever from connectable_observable

gchudnov:
Decoupled `connect_forever` from `connectable_observable`, please review.

Thank you,
Grigoriy


kirkshoop:
Thank you!

PR #341, opened 2017-01-28
decouple ref_count from connectable_observable

gchudnov:
Decoupled `ref_count` from `connectable_observable`
Put the include to the bottom of `rx-operators.hpp` as there is a dependency in [rx-synchronize](https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/subjects/rx-synchronize.hpp#L220)

Please review

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/rx-operators.hpp
+++ b/Rx/v2/src/rxcpp/rx-operators.hpp
@@ -96,7 +96,6 @@ public:
 }
 
 #include "operators/rx-lift.hpp"
-#include "operators/rx-ref_count.hpp"
 #include "operators/rx-subscribe.hpp"
 
 namespace rxcpp {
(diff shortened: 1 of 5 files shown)


Rx/v2/src/rxcpp/rx-operators.hpp
```diff
@@ -510,5 +516,6 @@ struct zip_tag {
 
 #include "operators/rx-multicast.hpp"
 #include "operators/rx-publish.hpp"
+#include "operators/rx-ref_count.hpp"
```
kirkshoop: So ref_count is also required to be in the default set.

kirkshoop:
🎉 By my count this was the last to be fully decoupled!

Thank you!

gchudnov:
You're welcome!

@kirkshoop 
Do you want to leave `subscribe` and `lift` _as-is_ or make it consistent with the rest of the operators, defining `some_tag` and providing `member_overload`?

kirkshoop:
I would like them to be consistent, that will make the docs consistent as well.

PR #344, opened 2017-01-31
doc update

### LINKED ISSUE #343: lift operator refactoring
@kirkshoop,

Just tried to [refactor](https://github.com/gchudnov/RxCpp/commit/d2463190703116b6bb6e5542f7bb34330cdd75cb) the `lift` operator, but it looks it doesn't fit exactly the new operator implementation pattern.

It would be great if you provided some implementation guidelines...

--
There is an explicit result type, e.g. ['int' in this test](https://github.com/gchudnov/RxCpp/commit/d2463190703116b6bb6e5542f7bb34330cdd75cb#diff-fe06b4f3e1014ed24c2b294e1eb192a1R177):

```
rxo::lift<int>(liftfilter([&invoked](int x) {...
           ^^^^
```
and I have to provide a [dummy variable](https://github.com/gchudnov/RxCpp/commit/d2463190703116b6bb6e5542f7bb34330cdd75cb#diff-3a14c614bdd8677f53f166b099fca77bR118) just for the sake of this type.
It works, but looks ugly... Not sure that is OK... 

Guess I can call `member_overload<lift_tag>::member` directly, but that _breaks_ the invocation pattern, where we call `observable_member(YYY_tag{},                *this, std::forward<AN>(an)...);` from the observable member function..

What do you think is the preferred solution?

gchudnov:
Updated docs for `lift` and `subscribe`
No other changes.
closes #343 

Please review.

Thank you,
Grigoriy

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-lift.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-lift.hpp
@@ -2,6 +2,17 @@
 
 #pragma once
 
+/*! \file rx-lift.hpp
+
+    \brief  takes any function that will take a subscriber for this observable and produce a subscriber.
+            this is intended to allow externally defined operators, that use make_subscriber, to be connected into the expression.
+
+    \tparam ResultType  the type of the emitted results.
+    \tparam Operator    the type of the operator.
+
+    \return An observable that emitting the items from its source.
+ */
+
 #if !defined(RXCPP_OPERATORS_RX_LIFT_HPP)
 #define RXCPP_OPERATORS_RX_LIFT_HPP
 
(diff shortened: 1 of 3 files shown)


gchudnov:
@kirkshoop 
btw, I noticed two `lift_if` methods, but haven't found any usage or tests. Are they required?

kirkshoop:
@gchudnov, I would not be surprised if they are not used. They were one attempt to delay the evaluation of a template method until it was used. They never worked well.

If they are not in use they should be deleted.

kirkshoop:
I cancelled the OSX build - apparently there is an outage on travisci

PR #345, opened 2017-02-01
add aliases and fix docs

kirkshoop:
added aliases for some operators
transform (map)
merge_transform (flat_map)
concat_transform (concat_map)
switch_on_error (on_error_resume_next)
accumulate (reduce)


kirkshoop:
@gchudnov, 

OSX machines on travis are overloaded, so I think that we should merge without them for now.

What do you think of this change? If you approve, you can merge.

gchudnov:
yea, sure, looks great!
let's merge!

PR #356, opened 2017-02-18
Rewrite repeat operator to handle 0 case correctly and not rely on magic numbers

elelel:
This rewrite addresses two isssues.
1. Rx operator repeat with 0 time argument should return an empty sequence. The reasons for that:
  * Rx API is designed based on traditional functional programming and the fact that Iterator/Observer patterns are functional duals. Repeat function with times argument in non-lazy languages with Iterator-implemented sequences returns empty sequence when called with 0 as times argument, it's natural for the Observer dual to behave simiraly
  * Using 0 as "magic number" breaks algorithmic generality, i.e. passing a dynamic counter variable to repeat as argument leads to unexpected result
  * Rx in other languages treat 0 that way (e.g. https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L8536 )

2. Get rid of dependency on dynamic state counter. Infinite repeat implementation should not be coupled with finite repeat, it should not make run-time checks for the finiteness flag

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-repeat.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-repeat.hpp
@@ -42,8 +42,8 @@ struct repeat_invalid : public rxo::operator_base<repeat_invalid_arguments<AN...
 template<class... AN>
 using repeat_invalid_t = typename repeat_invalid<AN...>::type;
 
-template<class T, class Observable, class Count>
-struct repeat : public operator_base<T>
+  template<class RepeatConcrete, class T, class Observable, class Count>
+struct repeat_base : public operator_base<T>
 {
     typedef rxu::decay_t<Observable> source_type;
     typedef rxu::decay_t<Count> count_type;
@@ -106,11 +106,11 @@ struct repeat : public operator_base<T>
                     },
                 // on_completed
                     [state]() {
-                        if (state->repeat_infinitely || (--state->remaining > 0)) {
-                            state->do_subscribe();
-                        } else {
-                            state->out.on_completed();
-                        }
+                      if (RepeatConcrete::on_completed_predicate(state->remaining)) {
+                        state->out.on_completed();
+                      } else {
+                        state->do_subscribe();
+                      }
                     }
                 );
             }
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@elelel__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

msotclas:
__@elelel__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, MSOTBOT;

kirkshoop:
Hi!

This looks great! 

This changes the behavior of repeat(0), do you have any ideas for how to stage the change in behavior?

If not this will require a major version bump.

Rx/v2/test/operators/repeat.cpp
```diff
@@ -56,6 +56,40 @@ SCENARIO("repeat, basic test", "[repeat][operators]"){
...
+                        | rxo::as_dynamic();
+                }
+            );
+
+            THEN("the output should be empty"){
+                auto required = std::vector<rxsc::test::messages<int>::recorded_type>();
+                auto actual = res.get_observer().messages();
+                REQUIRE(required == actual);
+            }
+
```
kirkshoop: please add a THEN clause to check the subscription lifetime.

Rx/v2/src/rxcpp/operators/rx-repeat.hpp
```diff
@@ -42,87 +42,144 @@ struct repeat_invalid : public rxo::operator_base<repeat_invalid_arguments<AN...
...
-                        } else {
-                            state->out.on_completed();
-                        }
-                    }
-                );
-            }
-        };
-
+      typedef state_type<values, Subscriber, T> state_t;
+      if (!initial_.completed_predicate()) {
```
kirkshoop: if repeat count is 0 should't this call on_completed like empty() does instead of calling nothing like never() does?

elelel: I thought about it a lot, the question is when should we call on_completed()? There were different versions, return on_completed() immediately. Or for example 0f0f23893a048cb7f104e31b069df6b55d4a1087 used to put on_completed() into sequence but only when the input sequence is fully consumed. So I left it for a future consideration. If you have a strong opinion when on_completed() should be called immedieately, let's put it there.

elelel:
> This changes the behavior of repeat(0), do you have any ideas for how to stage the change in behavior?
> If not this will require a major version bump.

We can add a wrapper to check for static repeat(0) calls and emit fatal/warning about the change. We can also add a wrapper that emits the warning and reverts to old behaviour.
I think version bump is inevitable as soon as we decide to let client use repeat(0) as supposed.

gchudnov:
Right, it looks that are breaking changes anyway, no matter whether you write some warning or not...
There should be a major version bump.
--
Another option is to create a new operator, like `repeatn` with the new behavior, but that should add more confusion...

kirkshoop:
okay, major version it is, Thanks!

Issue #355, opened 2017-02-17
Copy Avoidance


adsharma:
After digging into the micro benchmark linked here:

https://github.com/facebookincubator/iterlib/pull/25

and looking at some profiles, I suspect that there is a copy going on between the observable -> skip -> take. Is there a way to avoid the copy in this code?

```
auto values = rxcpp::observable<>::iterate(vec).skip(size/2).take(size/4);
```

adsharma:
I wonder if it's feasible to use iterator_traits::reference instead of value_type here:

https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/src/rxcpp/sources/rx-iterate.hpp#L159-L172

kirkshoop:
Hi!

rxcpp attempts to ensure minimal copies. Bugs fixes for excessive copies are welcome!

In this case, `iterate` is intentionally taking a copy of the vector when constructed and then copying each item during iteration. 

rxcpp is async by default. this means that `iterate` takes the vector by value so that a copy (or move) of the vector is invoked, so that when `subscribe` is called later, the data is still valid.

the vector copy would be avoided by passing a non-owning range to `iterate`.

`iterate` is __cold__. this means that `subscribe` can be called multiple times and should return the full set of values each time and that those values should not have been mutated.

I think that the per value copy could be removed using `iterator_traits::const_reference` to prevent mutation. perhaps you could provide a PR for that and we can test to see if there is any negative affect.

Thanks!

adsharma:
> the vector copy would be avoided by passing a non-owning range to iterate.

Thanks for the suggestion.

```
-    auto values = rxcpp::observable<>::iterate(vec).skip(size/2).take(size/4);
+    std::vector<ItemOptimized> sliced = vec | ranges::view::slice(size/2, size/2 + size/4);
+    auto values = rxcpp::observable<>::iterate(sliced);
```

speeds up things quite a bit by using range-v3. I would like to be able to write:

```
-    auto values = rxcpp::observable<>::iterate(vec).skip(size/2).take(size/4);
+    auto sliced = vec | ranges::view::slice(size/2, size/2 + size/4);
+    auto values = rxcpp::observable<>::iterate(sliced);
```

But the code doesn't compile. Is there a way to make a range-v3 view a source for rxcpp::observable without forcing a copy to a vector followed by iterate()? I suspect a new source would've to be written.

Re: PR for const_reference

[...]

PR #359, opened 2017-02-23
fix iterator_type::value_type error

kirkshoop:
#355 reported that range-v3 ranges failed due to an errant `&` in the
`iterator_type`


gchudnov:
looks great, merged!

PR #360, opened 2017-02-23
Retry: subscription lifetime; repeat: generic naming

elelel:
Add subscription tracking to retry operator to mange lifetime.

Change method naming in repeat operator implementation to match generically with retry

PR #363, opened 2017-02-25
Factor out commonalities between repeat and retry into a separate file

elelel:
This takes out the commonalities between rx-retry.hpp and rx-repeat.hpp into a single file to reduce the LOC number and prevent multiple place same-edits problems (like the one fixed in #360)


msotclas:
@elelel,
Thanks for having already signed the Contribution License Agreement. Your agreement was validated by Microsoft Open Technologies, Inc.. We will now review your pull request.
_Thanks,

Rx/v2/src/rxcpp/operators/impl-retry-repeat.hpp
```diff
@@ -0,0 +1,163 @@
...
+    
+*/
+
+#include "../rx-includes.hpp"
+
+namespace rxcpp {
+  namespace operators {
+    namespace detail {
+
+      namespace retry_repeat_common {
```
kirkshoop: I would prefer for the name of the file to match the name of the type or namespace that is introduces. This preference here does not block merge of this PR.

Rx/v2/src/rxcpp/operators/rx-repeat.hpp
```diff
@@ -25,166 +25,46 @@
 #define RXCPP_OPERATORS_RX_REPEAT_HPP
 
 #include "../rx-includes.hpp"
+#include "impl-retry-repeat.hpp"
```
kirkshoop: rxcpp was structured so that all include dependencies are expressed in `rx-includes.hpp` and a few of the `.hpp` it directly includes (sources, operators, etc..). RX-LITE changed the include structure to make the individual operator includes optional and move them later in the dependency graph.

some options are:

- leave this include structure as is. 
 - new pattern
- move this include to `rx-includes.hpp` 
 - will always be included even in RX-LITE when retry and repeat are not
- merge repeat and retry into a single `.hpp`
 - bleeds the implementation details to the user

Without RX-LITE I would choose differently, but I am leaning to keeping this include as is. What do you think?

Rx/v2/src/rxcpp/operators/impl-retry-repeat.hpp
```diff
@@ -0,0 +1,163 @@
...
+        template<class... AN>
+        struct invalid_arguments {};
+
+        template<class... AN>
+        struct invalid : public rxo::operator_base<invalid_arguments<AN...>> {
+          using type = observable<invalid_arguments<AN...>, invalid<AN...>>;
+        };
+        template<class... AN>
+        using invalid_t = typename invalid<AN...>::type;
```
kirkshoop: invalid should not be shared. These structs are used to provide the user with better information in compilation errors (the operator and the argument types that failed to resolve to a supported overload of that operator). `retry_repeat_common::invalid` is less helpful than `repeat::invalid`

Rx/v2/src/rxcpp/operators/rx-retry.hpp
```diff
@@ -21,165 +21,47 @@
...
-      }
-
-      source_type source;
-    };
-
-    infinite(source_type s) : initial_(std::move(s)) {
+          
+    template <typename State>
+    static inline void on_completed(State& state) {
+      // JEP: never appeears to be called?
```
kirkshoop: nothing to worry about, but I do not think that we need to keep this comment around anymore. :)

Rx/v2/src/rxcpp/operators/impl-retry-repeat.hpp
elelel: Could you be more specific on that? Do you have a naming convention approach? Should I rename the namespace into impl-retry-repeat, rename the file into retry_repeat_common, retry-repeat-common, rx-retry-repeat-common?

Rx/v2/src/rxcpp/operators/rx-repeat.hpp
elelel: I'm new to RxCpp dev and completely devoid of knowledge how the project is structured. Maybe there's some dev doc describing the coding style recommendations for the project? When structuring the files in this PR I based it solely on the facts from my C++ dev experience:
- When creating code of highly composable functionality there has to be a way to reuse common code in hierarchical way, otherwise code duplication occurs. In our case it is polymorphism through static inheritance. This implies deep dependency trees and that there should be some way to refer to single piece of code from two or more other pieces of codes (files). In other words, we have to include the base class from somewhere.
- This is a header-only library, so a flat include structure will sooner or later become toxic for compile time, so it's better not to be a single include dependencies file

Rx/v2/src/rxcpp/operators/impl-retry-repeat.hpp
kirkshoop: rx-retry-repeat-common.hpp follows the pattern for the rest of the files.

Rx/v2/src/rxcpp/operators/rx-repeat.hpp
kirkshoop: The code is the primary documentation :)
If the same feedback is required on several PRs from different authors I would find a way to document that. 
The developer manual was added by a contributor that wanted to record some things that he found helpful, please feel free to expand that with information you think is useful - or put it in code comments - or doxygen comments in the code - or find a better place and structure to store this kind of information.  

Thanks!

Rx/v2/src/rxcpp/operators/impl-retry-repeat.hpp
elelel: Fixed that. By the way, what does "rx" prefix stands for in the filenames? I thought it was to prefix something that came from Rx world, like operator names in this dir.

Rx/v2/src/rxcpp/operators/rx-retry-repeat-common.hpp
```diff
@@ -0,0 +1,153 @@
+#pragma once
+
+/*! \file impl-retry-repeat.hpp
```
kirkshoop: This will confuse doxygen..

kirkshoop:
Thank you!

PR #368, opened 2017-03-30
fixed https://github.com/Reactive-Extensions/RxCpp/issues/366

anatoly-spb:
I have no a rational explanation of my changes (may be optimizer does something wrong with a problem expression, I do not know), but now all tests passed with 32 bit release (with/without v140_xp):

`100% tests passed, 0 tests failed out of 65`

I found the minimal code which reproduces the problem with RelWithDebInfo build:

```
int main()
{
    auto sc = rxsc::make_test();
std::vector<rxcpp::notifications::recorded<std::shared_ptr<rxcpp::notifications::detail::notification_base<int> > > > v;
    auto xs = sc.make_hot_observable(v); // here is segmentation fault
    return 0;
}
```

which failed in depth of std::make_shared:

```
return rxt::testable_observable<T>(
        std::make_shared<hot_observable<T>>(state, create_worker(composite_subscription()), std::move(messages)));
``` 

After that I simplified the expression with local intermediate variables:

```
    auto worker = create_worker(composite_subscription());
    auto shared = std::make_shared<hot_observable<T>>(state, worker, std::move(messages));
    return rxt::testable_observable<T>(shared);
```

and the problem  disappeared.

### DIFF
--- a/Rx/v2/src/rxcpp/schedulers/rx-test.hpp
+++ b/Rx/v2/src/rxcpp/schedulers/rx-test.hpp
@@ -340,8 +340,9 @@ public:
 template<class T>
 rxt::testable_observable<T> test_type::make_hot_observable(std::vector<rxn::recorded<std::shared_ptr<rxn::detail::notification_base<T>>>> messages) const
 {
-    return rxt::testable_observable<T>(
-        std::make_shared<hot_observable<T>>(state, create_worker(composite_subscription()), std::move(messages)));
+    auto worker = create_worker(composite_subscription());
+    auto shared = std::make_shared<hot_observable<T>>(state, worker, std::move(messages));
+    return rxt::testable_observable<T>(shared);
 }
 
 template<class F>


msotclas:
This seems like a small (but important) contribution, so no Contribution License Agreement is required at this point. We will now review your pull request.
_Thanks,

gchudnov:
Thank you for the PR!

yes, it works, but let's spend some time investigating it. Just to be sure that we are not concealing the problem instead of fixing it.

Tried to debug it too, but it is still weird. It appears on /O1 optimization and the vector gets corrupted somehow(?) Need more time..

anatoly-spb:
If replace std::make_shared with std::shared_ptr<...>(new T...):  

  return rxt::testable_observable<T>(std::shared_ptr<hot_observable<T>>(new hot_observable<T>(state, create_worker(composite_subscription()), std::move(messages))));

there is also no problem.

kirkshoop:
I have asked the visual studio team to take a look, but no response yet. I would propose that we take this now. I will pursue a fix in visual studio.

Thanks!

gchudnov:
And my status on this problem is the following:
I wasn't able to identify the problem. :(
Initially i thought there is some issue in RxCPP but didn't find anything.
Still puzzled. Might be an issue in Visual Studio compiler(?)

--
@kirkshoop 
so, do you propose to merge this PR now and revisit it after the response from VS team?

kirkshoop:
Yes, I think that this is a compiler issue and we should merge this now and revisit once the VC team responds to me.

Thanks!

gchudnov:
Merged!
Thank you all, guys!

Issue #369, opened 2017-03-31
[Feature request] buffer with predicate which determines when buffer must be flushed


anatoly-spb:
Hello! 

I have simple example with character stream: 

```
  auto source = rxcpp::observable<>::iterate(
    std::initializer_list<char>{ '1', '2', '3', '$', '5', '$', '7', '8', '$' }
  );

```

I want to get std::vector<char> for each sub sequence terminated by '$'.  I know that there is  buffer operator ( http://reactivex.io/documentation/operators/buffer.html ) with count and time overloads. But what about buffer with predicate which determines when buffer must be flushed? 

Now I have the following workaround: 

```
  auto source = rxcpp::observable<>::iterate(
    std::initializer_list<char>{ '1', '2', '3', '$', '5', '$', '7', '8', '$' }
  );

  auto vector_stream = source
    | rxcpp::operators::flat_map([](char ch) {
    return rxcpp::observable<>::create<std::vector<char>>(
      [ch](rxcpp::subscriber<std::vector<char>> s) {
      static std::vector<char> a;
      if (ch == '$') {
        s.on_next(a);
        a.clear();
      }
      else {
        a.push_back(ch);
      }
    });
  });

  vector_stream
    .take(3)
    .subscribe([](const std::vector<char> &v) {
    std::cout << "consume vector: ";
    std::copy(v.begin(), v.end(), std::ostream_iterator<char>(std::cout, " "));
    std::cout << std::endl;
  });

```
I am newcomer, so may be you point me more correct and elegant way to solve my task. 

Thanks in advance,
Anatoly Shirokov

kirkshoop:
The `window_toggle` operator can be used to build this. In fact this is used in Readme.md to implement a variation of this exact problem.

```cpp
    // filter to last string in each line
    auto closes = strings |
        filter(
            [](const string& s){
                return s.back() == '\r';
            }) |
        Rx::map([](const string&){return 0;});

    // group strings by line
    auto linewindows = strings |
        window_toggle(closes | start_with(0), [=](int){return closes;});
```

Partially applied to this case:

```cpp
  auto source = rxcpp::observable<>::iterate(
    std::initializer_list<char>{ '1', '2', '3', '$', '5', '$', '7', '8', '$' }
  ) |
  // share
  publish() |
  ref_count();

    // filter to flushes
    auto flushes = source |
        filter(
            [](char c){
                return c == '$';
            }) |
        Rx::map([](const string&){return 0;});

[...]

PR #370, opened 2017-04-07
fix async lifetime bug in window_toggle

kirkshoop:
encountered in #369

This one took some time to debug. 😬


kirkshoop:
wrong fix.

PR #371, opened 2017-04-07
Kirkshoop/filterimprovement

kirkshoop:
proposed changes to #364 

@elelel, does this look okay?

> whitespace changes are due to enforcement of the `.editorconfig`


elelel:
Sure, I'm not informed enough to make suggestions here :)

Are we safe from the user calling on_next with a const rvalue directly?

kirkshoop:
Thanks!

Yes, one of the weird parts of perfect-forwarding in C++ is that since the `Value&& v` parameter is named, it is not an `R-Value` which is why `std::forward` must be used to recover the original type.

This means that this use of `as_const` in `on_next` will not ever be passed an `R-Value`.

kirkshoop:
@elel, thank you for the help in getting this done.

PR #377, opened 2017-05-15
Add CMake option to not build and run test and examples

per-gron:
It's super great that RxCpp has such a comprehensive set of tests and examples, but when I include RxCpp in my own (CMake based) project, I am not particularly interested in compiling them (which takes a significant amount of time, especially compared to the rest of my project) or running the tests with CTest.

CMake has a nice `EXCLUDE_FROM_ALL` option to its `add_subdirectory` command which makes it build only things that you depend on explicitly, but unfortunately it does not interact well with CTest. If I use `EXCLUDE_FROM_ALL` on RxCpp, then `ctest` fails because it attempts to run tests that were not built.

[As far as I know](http://stackoverflow.com/questions/23713650/cmake-exclude-tests-in-subdirectories) there is no sane way to disable CTest tests once they have been registered, and the idiomatic CMake approach seems to be to have an option like this that optionally disables tests and similar things. [Flatbuffers do this, for example](https://github.com/google/flatbuffers/blob/master/CMakeLists.txt#L7).

With this patch, I will be able to import RxCpp to my project like this:

    set(RXCPP_BUILD_TESTS OFF)
    add_subdirectory(rxcpp)

### DIFF
--- a/projects/CMake/CMakeLists.txt
+++ b/projects/CMake/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
 
 project(rxcpp VERSION 2.2.4 LANGUAGES C CXX)
 
+option(RXCPP_BUILD_TESTS "Enable the build of tests and samples." ON)
+
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 # define some folders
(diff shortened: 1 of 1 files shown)


anatoly-spb:
RxCpp is header only library, so you have not to build one at all inside your project. The only thing you should do is to customize include_directory:
`
INCLUDE_DIRECTORIES(${RXCPP_HOME} $ENV{RXCPP_HOME}) 

`

per-gron:
@anatoly-spb Okay! I'll close this then.

Thanks for looking at this PR!

PR #381, opened 2017-06-05
Add get_subscription() for subject

wutiantong:
I think this little code piece is harmless and meaningful.

Further explaining, there are two constructor of subject: 
```C++
    subject()
        : s(composite_subscription())
    {
    }
    explicit subject(composite_subscription cs)
        : s(cs)
    {
    }
```
to retrieve the auto-created subscription(in 1st constructor), we need a get_subscription() for subject.


kirkshoop:
I would expect that this subscription was already available by calling `get_subscriber().get_subscription()` but this is convenient.

Thanks!

PR #387, opened 2017-07-14
Fix timeout when no items are emitted

gchudnov:
Timeout triggers if no items are emitted.

Please review.

Thank you,
Grigoriy


kirkshoop:
Thanks!

Issue #388, opened 2017-07-20
rx-util.hpp should stop calling std::unexpected()


StephanTLavavej:
[rx-util.hpp line 696](https://github.com/Reactive-Extensions/RxCpp/blob/201dd99b6494564c9e8990d0a5c32b23431128d4/Rx/v2/src/rxcpp/rx-util.hpp#L696) says:

```
    ~unwinder()
    {
        if (!!function)
        {
            try {
                (*function)();
            } catch (...) {
                std::unexpected();
            }
        }
    }
```

`std::unexpected()` was deprecated in C++11 and removed in C++17. In VS 2017's second toolset update (not 15.3, the first toolset update), activating C++17 mode with `/std:c++17` will remove `std::unexpected()` by default.

To conform to C++17, this call should be changed. I recommend calling `std::terminate()` instead.

kirkshoop:
ok, Thanks!

PR #389, opened 2017-07-20
std::unexpected has been deprecated

kirkshoop:
fix for #388


gchudnov:
Looks great, thank you!

PR #391, opened 2017-07-22
Add 'include' to the install path of the headers

petoknm:
When installing the headers (as far as I see there are no library files) into the system using `make install` installs them into `/usr/local/rxcpp/...` which is not a standard place for header files...  This PR fixes that by installing them into something more like `/usr/local/include/rxcpp/...`. Of course the `/usr/local` is just my default `CMAKE_INSTALL_PREFIX` which can be changed to whatever people like. This will allow me to package rxcpp more easily into and AUR package which I'll hopefully soon make because I didnt find any out there.


gchudnov:
Good catch, thank you!

PR #399, opened 2017-09-11
Fix clang-trunk -Wunused-lambda-capture warnings

ericniebler:
By default, RxCpp fails to build with clang-trunk due to `-Werror,-Wunused-lambda-capture` warnings. Remove unused lambda captures.


Rx/v2/test/operators/pairwise.cpp
```diff
@@ -21,7 +21,7 @@ SCENARIO("pairwise - enough items to create pairs", "[pairwise][operators]") {
         WHEN("taken pairwise") {
 
             auto res = w.start(
-                [xs, &invoked]() {
```
kirkshoop: according to visual studio this makes `invoked` decl on line 10 unreferenced. :)

kirkshoop:
Thanks!

PR #400, opened 2017-09-15
Fix rxcpp::observable<>::from<T>() when T is non-default constructable

lebdron:
The following code does not compile with AppleClang 8.1.0.8020042 due to implementation of `std::array` internally storing array with one element even if it is declared as `std::array<T, 0>`.
```C++
class foo {
    foo(int) {}
};
auto o = rxcpp::observable::empty<foo>();
```
This PR fixes this issue by using `std::initializer_list<T>` instead `of std::array<T, 0>`.


kirkshoop:
Thank you!

PR #401, opened 2017-09-15
Fix rxcpp::observable<>::range<T>() when T is unsigned type

lebdron:
The following code does not compile since `std::abs` is not defined for unsigned types. Moreover, subtraction of two unsigned values may lead to unwanted behaviour.
```C++
auto a = 1u, b = 2u;
rxcpp::observable<>::range(a, b).subscribe([] {});
```
This PR fixes the issue with basic `std::max() - std::min()` approach, which is applicable for both signed and unsigned values.


kirkshoop:
Thank you!

PR #417, opened 2017-12-06
Add rx-merge-delay-error operator

nitrram:
The operator and its test is copy-pasted and modified accordingly from `merge`
It should behave the same way as JavaRx `mergeDelayError()` in the terms of semantics

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-merge_delay_error.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-merge_delay_error.hpp
@@ -1,0 +1,302 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+#pragma once
+
+/*! \file rx-merge_delay_error.hpp
+
+        \brief For each given observable subscribe.
+                   For each item emitted from all of the given observables, deliver from the new observable that is returned.
+				   The first error to occure is hold off until all of the given non-error-emitting observables have finished their emission.
+
+                   There are 2 variants of the operator:
+                   - The source observable emits nested observables, nested observables are merged.
+                   - The source observable and the arguments v0...vn are used to provide the observables to merge.
+
+        \tparam Coordination  the type of the scheduler (optional).
+        \tparam Value0	... (optional).
+        \tparam ValueN	types of source observables (optional).
+
+        \param	cn		the scheduler to synchronize sources from different contexts (optional).
+        \param	v0		... (optional).
+        \param	vn		source observables (optional).
+
+        \return																											 Observable that emits items that are the result of flattening the observables emitted by the source observable.
+
+        If scheduler is omitted, identity_current_thread is used.
+
+        \sample
+        \snippet merge_delay_error.cpp threaded implicit merge sample
+        \snippet output.txt threaded implicit merge sample
+
+        \sample
+        \snippet merge_delay_error.cpp implicit merge sample
+        \snippet output.txt implicit merge sample
+
+        \sample
+        \snippet merge_delay_error.cpp merge sample
+        \snippet output.txt merge sample
+
+        \sample
+        \snippet merge_delay_error.cpp threaded merge sample
+        \snippet output.txt threaded merge sample
+*/
+
+#if !defined(RXCPP_OPERATORS_RX_MERGE_DELAY_ERROR_HPP)
+#define RXCPP_OPERATORS_RX_MERGE_DELAY_ERROR_HPP
+
+#include "rx-merge.hpp"
+
+#include "../rx-composite_exception.hpp"
+
+namespace rxcpp {
+
+namespace operators {
+
+namespace detail {
+
+template<class T, class Observable, class Coordination>
+struct merge_delay_error
+        : public operator_base<rxu::value_type_t<rxu::decay_t<T>>>
+{
+        //static_assert(is_observable<Observable>::value, "merge requires an observable");
+        //static_assert(is_observable<T>::value, "merge requires an observable that contains observables");
+
+        typedef merge_delay_error<T, Observable, Coordination> this_type;
+
+        typedef rxu::decay_t<T> source_value_type;
+        typedef rxu::decay_t<Observable> source_type;
+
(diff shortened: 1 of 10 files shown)


Rx/v2/src/rxcpp/operators/rx-merge_delay_error.hpp
```diff
@@ -0,0 +1,302 @@
...
+                                );
+
+                                auto selectedSinkInner = state->coordinator.out(sinkInner);
+                                selectedSource.subscribe(std::move(selectedSinkInner));
+                        },
+                // on_error
+                        [state](std::exception_ptr e) {
+			                if(--state->pendingCompletions == 0) {
+			                    state->out.on_error(e);
+			                } else {
```
kirkshoop: perhaps when the last stream terminates in an error the error should be added to the `composite_exception` and then the `composite_exception` is sent to `out.on_error`?

Rx/v2/src/rxcpp/operators/rx-merge_delay_error.hpp
```diff
@@ -0,0 +1,302 @@
...
+                                        innercs,
+                                // on_next
+                                        [state, st](value_type ct) {
+                                                state->out.on_next(std::move(ct));
+                                        },
+                                // on_error
+                                        [state](std::exception_ptr e) {
+                                                if(--state->pendingCompletions == 0) {
+                                                        state->out.on_error(e);
+                                                } else {
```
kirkshoop: 'composite_exception` again?

Rx/v2/src/rxcpp/rx-composite_exception.hpp
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+#pragma once
+
+#ifndef _MSC_VER
+#define NOEXCEPT noexcept
+#else
+#define NOEXCEPT
+#endif
+
```
kirkshoop: this should be in rx-includes.hpp and should be specific to MSVC version (I know that recent versions support noexcept) and should have an unambiguous name to prevent conflicts `RXCPP_NOEXCEPT`

Rx/v2/test/operators/merge_delay_error.cpp
```diff
@@ -0,0 +1,390 @@
...
+                .count();
+
+            auto finish = clock::now();
+            auto msElapsed = duration_cast<milliseconds>(finish.time_since_epoch()) -
+                   duration_cast<milliseconds>(start.time_since_epoch());
+            std::cout << "merge serial ranges : " << n << " subscribed, " << c << " emitted, " << msElapsed.count() << "ms elapsed " << std::endl;
+        }
+    }
+}
+
```
kirkshoop: it is not required to keep the [perf] tagged tests above, since the error handling perf is not being tested.

Rx/v2/test/operators/merge_delay_error.cpp
```diff
@@ -0,0 +1,390 @@
...
+                    on.next(310, 103),
+                    on.next(320, 104),
+                    on.next(320, 305),
+                    on.next(410, 105),
+                    on.next(430, 107),
+                    on.error(440, ex)
+                });
+                auto actual = res.get_observer().messages();
+                REQUIRE(required == actual);
+            }
```
kirkshoop: it would be nice to be able to check that the exception list had the correct exceptions stored.. but I think that would require a lot of work in the test infrastructure. something for the future.

Rx/v2/src/rxcpp/operators/rx-merge_delay_error.hpp
```diff
@@ -0,0 +1,302 @@
...
+        \snippet output.txt implicit merge sample
+
+        \sample
+        \snippet merge_delay_error.cpp merge sample
+        \snippet output.txt merge sample
+
+        \sample
+        \snippet merge_delay_error.cpp threaded merge sample
+        \snippet output.txt threaded merge sample
+*/
```
kirkshoop: These examples need to be defined in a new file next to https://github.com/Reactive-Extensions/RxCpp/blob/master/Rx/v2/examples/doxygen/merge.cpp

You might want to have an example that demonstrates the `composite_error`

Rx/v2/test/operators/merge_delay_error.cpp
nitrram: yes, I agree - I also spotted that the test does not cover the actual exception assertation. I'll take a look at that later

Rx/v2/src/rxcpp/operators/rx-merge_delay_error.hpp
nitrram: I added those doxygen examples:
* https://github.com/nitrram/RxCpp/blob/merge_error_delay/Rx/v2/examples/doxygen/merge_delay_error.cpp
* https://github.com/nitrram/RxCpp/blob/merge_error_delay/Rx/v2/examples/doxygen/composite_exception.cpp

-- hopefully sufficiently enough

Yet if there is something missing, let me know.

kirkshoop: looks great! I did not see these added to https://github.com/Reactive-Extensions/RxCpp/blob/master/projects/doxygen/CMakeLists.txt 

did I miss that?

nitrram: Ah, my bad! I forgot that. I should also change the documentation at rx-merge_delay_error. cpp so that it matches the tests which I'm afraid does not at the moment. Unfortunately I'll be at a computer not sooner than on Sunday. But I'll fix it right then. Thanks for the review and patience

nitrram: Ok, so eventually I managed that. I added those 2 samples among the others. The documentation turned to be fine to me after all. When it's approved, shall I do something about it (squash commits...), before it gets merged?

kirkshoop:
Thank you so much for adding the `merge_delay_error()` operator!

PR #442, opened 2018-05-10
Remove unused variable and captures in doxygen/buffer.cpp

lebdron:
Fix build error with Clang 6.0.0
```
/Users/andrei/CLionProjects/RxCpp/Rx/v2/examples/doxygen/buffer.cpp:174:14: error: lambda capture 'start' is not used [-Werror,-Wunused-lambda-capture]
            [start](std::vector<long> v){
             ^
/Users/andrei/CLionProjects/RxCpp/Rx/v2/examples/doxygen/buffer.cpp:195:14: error: lambda capture 'start' is not used [-Werror,-Wunused-lambda-capture]
            [start](std::vector<long> v){
```


kirkshoop:
Looks great!

PR #443, opened 2018-05-10
Fix exit_recursed_scope_type dtor cleaning requestor early

lebdron:
Because of disabled RVO, `exit_recursed_scope_type` destructor was called on return statement in `recursed_scope_type::reset`, making the call ineffective. This change allows to use it as intended when RVO is disabled.

### DIFF
--- a/Rx/v2/src/rxcpp/rx-scheduler.hpp
+++ b/Rx/v2/src/rxcpp/rx-scheduler.hpp
@@ -458,11 +458,18 @@ class schedulable : public schedulable_base
         public:
             ~exit_recursed_scope_type()
             {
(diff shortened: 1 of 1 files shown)


kirkshoop:
Thank you! 

The `shared_ptr<>` will undo the performance provided by the recursed pattern. Another approach to try is to make `exit_recursed_scope_type` moveable and only set `that->requestor = nullptr` when `that != nullptr`. The move constructor would make sure that only the destination had `that != nullptr`

lebdron:
@kirkshoop Thank you for clarification. I have applied the changes as you have described.

kirkshoop:
Looks great!

PR #449, opened 2018-07-08
Remove static from observe_on_run_loop

lebdron:
The change allows to use different run loops during the program execution.


kirkshoop:
Thank you for the PR! Nice catch!

PR #456, opened 2018-08-11
Add support for compiling rxcpp with -fno-exceptions

iam:
Shall we continue our discussion on how we can upstream this now that I have rebased everything to tip-of-tree ?

```
100% tests passed, 0 tests failed out of 67

Total Test time (real) =   9.03 sec
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -168,7 +168,7 @@ before_script:
   # generate build
   ############################################################################
   - cd ${TRAVIS_BUILD_DIR}
-  - cmake . -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DRX_USE_EXCEPTIONS=$EXCEPTIONS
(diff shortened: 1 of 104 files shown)


kirkshoop:
looks great! the CI builds seem to need some minor changes. vc will need `[[noreturn]]` to be macroized. the doxygen examples have a few minor issues.

iam:
That sounds straightforward enough to fix.

Clarifying questions:

- Did you want me to plumb through the `fno-except/fno-rtti` up to the CMakeFiles somehow as an extra flag? 
- What about adding presubmit support for this new mode?

(Note that I am unfamiliar with CMake best practices, so I would be glad to hear of a specific suggestion if you have one in mind).

kirkshoop:
Good points. Regressions will be much less likely is there is a cmake option and a ci pass for this. I’ll add something for those if you don’t get to it first. :)

I am certainly not a cmake proficient but I did recently add some cmake options to change compiler flags in the facebookresearch/pushmi repo that could be a start.

PR #464, opened 2018-10-16
Fix doc target missing declarations errors

gsauthof:
Building the docs (e.g. with `ninja doc`) failed with GCC (e.g. GCC 7.3
under Fedora 27) with some missing declaration errors.

Example:

    ../Rx/v2/examples/doxygen/main.cpp:7:13: error: no previous
    declaration for ‘std::__cxx11::string get_pid()’
    [-Werror=missing-declarations]

Thus, this change adds a shared declaration for `get_pid()` that is
included by all users/the file where it's defined such that
accidental deviation in the signature are immediately noticed (as
compile error).

Similarly, the visibility of the example `less()` is changed to static
as it's only locally used.

With those changes the doc target succeeds again.

### DIFF
--- a/Rx/v2/examples/doxygen/amb.cpp
+++ b/Rx/v2/examples/doxygen/amb.cpp
@@ -30,7 +30,7 @@ SCENARIO("implicit amb sample"){
     printf("//! [implicit amb sample]\n");
 }
 
-std::string get_pid();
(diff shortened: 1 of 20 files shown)


kirkshoop:
Thanks!

Issue #473, opened 2019-02-01
captured connectable_observable varible cannot call connect method


ZichaoNickFox:
The question descripted in 
https://stackoverflow.com/questions/54472116/compile-error-in-reactive-extensions-for-cpp

I think this is caused by connectable_observable::connect is not a const method. When o is captured it turns to const variable, calling a non const connect method, the error occured.

I would send a merge request, please think about this.

kirkshoop:
Hi! Thanks for contributing. 

I would normally make the lambda mutable. Connect is not const becase it felt like a mutating operation.

Do you think it should be const because of the capture default or because the operation itself makes sense on a const object?

ZichaoNickFox:
Hi!

I did not know lambda mutable before. After study some, I add mutable specifier in the on_completed function as follow.
```c++
void publish_example() {
	auto o = rxcpp::observable<>::create<int>([](rxcpp::subscriber<int> s) {
		s.on_next(0);
		s.on_next(1);
	}).publish();
	o.subscribe([](int v) {printf("%d", v); });
	rxcpp::observable<>::timer(std::chrono::microseconds(2000)).subscribe(
		[](const int i) {},
		[](const std::exception_ptr& e) {},
		[o]() mutable {
		o.connect();
	} 
	);
}
```
Howere, a new compiler error occured as follow.
```c++
1>c:\users\liuzichao\source\repos\learnrx\learnrx\rxcpp\rx-observer.hpp(339): error C3848: expression having type 'const publish_example::<lambda_b6ccf59df0a8b3ff4481e4f6c25a801e>' would lose some const-volatile qualifiers in order to call 'void publish_example::<lambda_b6ccf59df0a8b3ff4481e4f6c25a801e>::operator ()(void)'
1>c:\users\liuzichao\source\repos\learnrx\learnrx\rxcpp\rx-observer.hpp(338): note: while compiling class template member function 'void rxcpp::observer<T,rxcpp::detail::stateless_observer_tag,_Ty,publish_example::<lambda_6d4024107010781f144cb4559793967e>,publish_example::<lambda_b6ccf59df0a8b3ff4481e4f6c25a801e>>::on_completed(void) const'
1>        with
1>        [
1>            T=long,
1>            _Ty=publish_example::<lambda_f2575ffc18258541d5828754a042c73b>
1>        ]
[...]

PR #474, opened 2019-02-01
fix issue #473

ZichaoNickFox:


kirkshoop:
we found another change to fix this - closing this change. thanks!

Issue #478, opened 2019-02-11
cmake fails to generate makefiles


iam:
Running this command on some Ubuntu systems:
```
cmake -G"Unix Makefiles" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=RelWithDebInfo -B. ../CMake
```
(from README.md "Linux --- Clang" instructions)

Gives the following error message:
```
    "/usr/bin/clang++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /mnt/ssd3/workspace/github/iam/rxcpp/projects/build/CMakeFiles/CMakeTmp
    
    Run Build Command:"/usr/bin/make" "cmTC_861e7/fast"
    /usr/bin/make -f CMakeFiles/cmTC_861e7.dir/build.make CMakeFiles/cmTC_861e7.dir/build
    make[1]: Entering directory '/mnt/ssd3/workspace/github/iam/rxcpp/projects/build/CMakeFiles/CMakeTmp'
    Building CXX object CMakeFiles/cmTC_861e7.dir/testCXXCompiler.cxx.o
    /usr/bin/clang++     -o CMakeFiles/cmTC_861e7.dir/testCXXCompiler.cxx.o -c /mnt/ssd3/workspace/github/iam/rxcpp/projects/build/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    Linking CXX executable cmTC_861e7
    /usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_861e7.dir/link.txt --verbose=1
    /usr/bin/clang++       -rdynamic CMakeFiles/cmTC_861e7.dir/testCXXCompiler.cxx.o  -o cmTC_861e7 
    /usr/bin/ld: cannot find -lstdc++
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
    CMakeFiles/cmTC_861e7.dir/build.make:97: recipe for target 'cmTC_861e7' failed
    make[1]: *** [cmTC_861e7] Error 1
    make[1]: Leaving directory '/mnt/ssd3/workspace/github/iam/rxcpp/projects/build/CMakeFiles/CMakeTmp'
    Makefile:126: recipe for target 'cmTC_861e7/fast' failed
    make: *** [cmTC_861e7/fast] Error 2
```

[...]

PR #479, opened 2019-02-11
README: fix cmake generation snippet for clang/linux

iam:
clang should be told to use libc++, otherwise it will
try to use libstdc++ which may be missing on some systems.

Resolves: #478


iam:
I apologize in advance for not updating Readme.html, I wasn't sure what the command to rebuild it was?

kirkshoop:
this is great, thanks!

Issue #477, opened 2019-02-11
'make all / make RxCpp' targets are broken


iam:
```
iam@invoker:/mnt/ssd3/workspace/github/iam/rxcpp/projects/build$ make -j48

[ 22%] Linking CXX shared library libRxCpp.so
[ 23%] Building CXX object test/CMakeFiles/rxcpp_test_distinct.dir/operators/distinct.cpp.o
[ 25%] Building CXX object test/CMakeFiles/rxcpp_test_on_error_resume_next.dir/operators/on_error_resume_next.cpp.o
[ 26%] Building CXX object test/CMakeFiles/rxcpp_test_start_with.dir/operators/start_with.cpp.o
[ 26%] Building CXX object test/CMakeFiles/rxcpp_test_observer.dir/subscriptions/observer.cpp.o
[ 26%] Building CXX object test/CMakeFiles/rxcpp_test_sample.dir/operators/sample.cpp.o
[ 27%] Building CXX object test/CMakeFiles/rxcpp_test_skip_last.dir/operators/skip_last.cpp.o
[ 27%] Building CXX object test/CMakeFiles/rxcpp_test_window.dir/operators/window.cpp.o
[ 29%] Building CXX object test/CMakeFiles/rxcpp_test_buffer.dir/operators/buffer.cpp.o
[ 30%] Building CXX object test/CMakeFiles/rxcpp_test_skip_until.dir/operators/skip_until.cpp.o
/usr/bin/ld: cannot find -lstdc++
clang: error: linker command failed with exit code 1 (use -v to see invocation)
CMakeFiles/RxCpp.dir/build.make:68: recipe for target 'libRxCpp.so' failed
make[2]: *** [libRxCpp.so] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/RxCpp.dir/all' failed
make[1]: *** [CMakeFiles/RxCpp.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

with verbose
```
$> VERBOSE=1 make -j32 RxCpp

/usr/bin/clang++ -fPIC -O2 -g -DNDEBUG  -shared -Wl,-soname,libRxCpp.so -o libRxCpp.so  
/usr/bin/ld: cannot find -lstdc++
clang: error: linker command failed with exit code 1 (use -v to see invocation)
CMakeFiles/RxCpp.dir/build.make:68: recipe for target 'libRxCpp.so' failed
[...]

PR #480, opened 2019-02-11
cmake: Fix 'make' generator targets

iam:
Remove 'RxCpp' library target when it's used outside of
Visual Studio. This was supposed to be a virtual target,
not an actual one that builds libRxCpp.so.

The build never worked because it was missing setting the
compiler/linker flags from shared.cmake

Resolves: #477


iam:
Maybe flake? Didn't see this CI failure on other PRs.

```
E: Unable to locate package g++-4.9
E: Couldn't find any package by glob 'g++-4.9'
E: Couldn't find any package by regex 'g++-4.9'
```

kirkshoop:
Yes, that was a flakey build. Thanks!

Issue #475, opened 2019-02-08
Possible data race in composite_subscription?


iam:
`composite_subscription_inner` looks like it might have an incorrect implementation of double-checked locking ?

https://github.com/ReactiveX/RxCpp/blob/7930ccc3f5eca55c5866fc7b1d2476f46b8cec17/Rx/v2/src/rxcpp/rx-subscription.hpp#L245

```c++
        inline weak_subscription add(subscription s) {
            if (!issubscribed) {
                s.unsubscribe();
            } else if (s.is_subscribed()) {
                std::unique_lock<decltype(lock)> guard(lock);          // <<<<<<<<<<<<<
                subscriptions.insert(s);
            }
            return s.get_weak();
        }
```

After acquiring the guard lock, issubscribed could already be false. Subsequently this races against either `clear` or `unsubscribe`. 

```
   Thread 1                                Thread 2
-----------------------------------------------------------
    add(s)            
                                       clear() or unsubscribe()
                                        lock
                                           call unsubscribe for all
                                        unlock
        block-for-lock
  -->  insert into set
        unlock
```

Likely correct code:

```c++
        inline weak_subscription add(subscription s) {
            if (!issubscribed) {
                s.unsubscribe();
            } else if (s.is_subscribed()) {
                std::unique_lock<decltype(lock)> guard(lock);
  /*+*/         if (!issubscribed.load()) {
  /*+*/           s.unsubscribe();
  /*+*/          } else {
                   subscriptions.insert(s);
                }
            }
            return s.get_weak();
        }
```

This potential issue seems repeated in most lock usages (`add`, `remove`, `clear`) except `unsubscribe` which is ok because of the exchange.

`remove()` also seems to have an additional data race with `w`:

```c++
        inline void remove(const weak_subscription& w) {
          if (issubscribed && !w.expired()) {           
            auto s = subscription::lock(w);                           /// <<< (2) this could already be expired
            std::unique_lock<decltype(lock)> guard(lock);  /// <<< (1) issubscribed could be false
            subscriptions.erase(s);                                      /// calls erase on moved-from set, could be benign
          }
        }
```

and calling 'subscription::lock' on an expired subscription would then call std::terminate (which btw, seems like a strange API decision) but at least the program would crash immediately instead of having data corruption.

-------
Another layer of possibly unspecified behaior:

The data race in expressed here in that std::set 'subscriptions' could be used (erase, insert, etc) even after it is moved-from in `erase`. 

[...]

PR #481, opened 2019-02-11
rxcpp: Fix data race in composite_subscription

iam:
composite_subscription_inner had missing checks which could lead to add/remove/clear
racing against unsubscribe.

(See the issue for more details).

Fixes: #475

### DIFF
--- a/Rx/v2/src/rxcpp/rx-subscription.hpp
+++ b/Rx/v2/src/rxcpp/rx-subscription.hpp
@@ -117,6 +117,14 @@ private:
             std::terminate();
         }
     }
+
+    explicit subscription(std::shared_ptr<base_subscription_state> s)
+        : state(std::move(s))
+    {
+        if (!state) {
+            std::terminate();
+        }
+    }
 public:
 
     subscription()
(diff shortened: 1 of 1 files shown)


Rx/v2/src/rxcpp/rx-subscription.hpp
```diff
@@ -242,41 +248,107 @@ class composite_subscription_inner
...
         inline void remove(weak_subscription w) {
             if (issubscribed && !w.expired()) {
+                // keep-alive for duration of rest of call
+                auto strong_subscription = w.lock();
+                if (!strong_subscription) {
+                  // non-expired weak_ptr could be subsequently expired.
+                  // subscription::lock(w) is not allowed to be called on
+                  // an expired pointer.
+                  return;
+                }
```
kirkshoop: the subscription constructor must terminate on a bad weak ptr because a subscription object must be valid (holding a strong reference to the shared state). 

My suggestion here would be to add a new overload of this `subscription` constructor:
  https://github.com/ReactiveX/RxCpp/blob/fd5f415b7ade514a2e04a5789d94e280203b295d/Rx/v2/src/rxcpp/rx-subscription.hpp#L113-L119

The new overload should take `std::shared_ptr<base_subscription_state>` 
The new overload should terminate if the pointer is null.

Then replace 
  `auto s = subscription::lock(w);`
with
  `auto s = subscription{std::move(strong_subscription)};`

kirkshoop:
This looks great! I love the comments!

Rx/v2/src/rxcpp/rx-subscription.hpp
iam: Ok, I added new constructor as you suggested and exposed it via a `rxu::maybe<subscription> maybe_lock(weak)` function.

kirkshoop:
Looks great!

Issue #482, opened 2019-02-11
travis CI failures [stack.nl is not accessible]


iam:
`[](https://travis-ci.org) ** continuous-integration/travis-ci/pr**` target appears to fail

Here's an example:
https://travis-ci.org/ReactiveX/RxCpp/jobs/491863729

```
The command "wget --quiet -O - http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.11.linux.bin.tar.gz" failed 3 times.
```

Visiting ftp.stack.nl, we get a DNS name resolution failure. Their site states that maybe they aren't offering these public services anymore?

https://www.stack.nl/notice.html

Should we upload `doxygen-1.8.11.linux.bin.tar.gz` somewhere else (perhaps github itself) and fix the scripts?

kirkshoop:
Thanks for this. I merged a different fix for this issue.

PR #476, opened 2019-02-11
Small fixes.

besser82:
### Fix 'Wcatch-value'.
This warning is reported by GCC 9.

### Fix url to Doxygen binary tarball.
The files hosted on SourceForge are not going to fade away.

PR #483, opened 2019-02-11
travis: Change url of doxygen binary

iam:
ftp.stack.nl is no longer accessible,
this is causing the travis CI (for clang) to fail.

Fixes: #482


iam:
Looks like this was already fixed in #476 but not merged yet?

kirkshoop:
thank you! yes, this is fixed in #476, which I just merged

Issue #484, opened 2019-02-13
Diamond graphs for observable aren't composable?


iam:
When trying to design an algorithm that requires turning a regular rx chain into a diamond (imagine there's no multi-threading necessary here), there seems to be a lack of composability:

```
     A
   /   \
  B     C
   \   /
     D
     |
     E
```
(A, B, C, D, E are some linear chains)

The (only working) approach I could figure out to this would look like:

```
auto A = ... | publish();
auto B = A | ...;
auto C = A | ...;
auto D = B | merge(C);
auto E = D | ...;

E | subscribe(...);
A.connect();                    // <<<<< not composable
```

The problem is that calling `connect` doesn't compose, so the whole graph can't just be expressed as an `observable<typeof(E)>`.

What's the best practice to solve this? Should we change some operators to make this easier?

---------------

RxJava has `autoConnect` which rxcpp seems to lack.

`ref_count` would almost work except it calls `connect` too soon (when `B` is subscribed, but before waiting for `C`).

--------------

Possible solutions:

- Add `auto_connect` to `connectable_observable` (problem: it doesn't seem to call 'unsubscribe' ever in `RxJava`. Not sure if this is acceptable?)

```
auto A = ... | publish() | auto_connect(2);
auto B = A | ...;
auto C = A | ...;
auto D = B | merge(C);
auto E = D | ...;

E | subscribe(...);
```

- Change `ref_count` to take a threshold value
          `ref_count(N)`:   call connect() when reaching >=N subscribers.
                                    unsubscribe from source when reaching 0 subscribers
  (Today's behavior is simply N=1).

```
auto A = ... | publish() | ref_count(2);
auto B = A | ...;
auto C = A | ...;
auto D = B | merge(C);
auto E = D | ...;

E | subscribe(...);
```

- Change `ref_count` to take another observable
          `ref_count(Obs)`:   subscribes/unsubscribes to its source as normal. also call Obs.connect() when reaching >=1 subscribers.
  (Today's behavior is simply Obs=source).

```
auto A = ... | publish();
auto B = A | ...;
auto C = A | ...;
auto D = B | merge(C) | ref_count(A);
auto E = D | ...;

E | subscribe(...);
```
- Add some kind of new operator to avoid `publish`?

```
auto A = <src> | fork(2);
auto B = A | ...;
auto C = A | ...;
auto D = B | merge(C);
auto E = D | ...;

E.subscribe(...);
/*
  calls D.subscribe which:
[...]

PR #485, opened 2019-02-15
operators: Add ref_count(other) operator overload.

iam:
The existing `connectable_observable.ref_count()` operator calls
connect on the source when it's subscribed to.

Generalize this by allowing an optional parameter `other`, i.e.
`observable.ref_count(connectable_observable other)` to be used as the
connect target.

Useful for implementing diamond graphs while retaining composability:

```
     A
   /   \
  B     C
   \   /
     D
     |
     E

auto A = ... | publish();
auto B = A | ...;
auto C = A | ...;
auto D = B | merge(C) | ref_count(A);
auto E = D | ...;

E | subscribe(...);
```

Resolves: https://github.com/ReactiveX/RxCpp/issues/484

### DIFF
--- a/Rx/v2/test/operators/publish.cpp
+++ b/Rx/v2/test/operators/publish.cpp
@@ -2,6 +2,8 @@
 #include <rxcpp/operators/rx-publish.hpp>
 #include <rxcpp/operators/rx-connect_forever.hpp>
 #include <rxcpp/operators/rx-ref_count.hpp>
+#include <rxcpp/operators/rx-map.hpp>
+#include <rxcpp/operators/rx-merge.hpp>
 
 
 SCENARIO("publish range", "[!hide][range][subject][publish][subject][operators]"){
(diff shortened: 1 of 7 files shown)


Rx/v2/test/operators/publish.cpp
```diff
@@ -38,6 +40,104 @@ SCENARIO("publish range", "[!hide][range][subject][publish][subject][operators]"
     }
 }
 
+SCENARIO("publish ref_count", "[range][subject][publish][ref_count][operators]"){
```
kirkshoop: I notice that these tests do not use the test scheduler or the on.next(time, value), etc.. utility. I would like to see these tests follow the normal pattern.

iam: I was a bit confused by the functionality of the test scheduler. In this case, I wanted to exercise that the 'same-thread' (i.e. no scheduler) is doing the right thing. Was that the wrong thing to test?

What is the recommended pattern to test that?

kirkshoop: the test scheduler does all work on the current thread. it is a good way to test this. it also records the marble values and when they happen.

iam: I believe I made the test changes requested, see 747366b

PR #493, opened 2019-04-07
Fix synchronize lock and recursion data race

### LINKED ISSUE #491: subjects::synchronize lock with current_thread or immediate scheduler
```C++
rxcpp::composite_subscription lifetime;
rxcpp::subjects::synchronize<int, rxcpp::identity_one_worker> sub(
    rxcpp::identity_current_thread(), lifetime);

sub.get_subscriber().on_next(1);
lifetime.unsubscribe();
```

This code locks in `drain_queue` https://github.com/ReactiveX/RxCpp/blob/7c79c18ab252eb37bcf122c75070f8d557868c08/Rx/v2/src/rxcpp/subjects/rx-synchronize.hpp#L60
because the lock in `on_next` is not unlocked https://github.com/ReactiveX/RxCpp/blob/7c79c18ab252eb37bcf122c75070f8d557868c08/Rx/v2/src/rxcpp/subjects/rx-synchronize.hpp#L107

Should it be safe to call `unlock` before these lines?
https://github.com/ReactiveX/RxCpp/blob/7c79c18ab252eb37bcf122c75070f8d557868c08/Rx/v2/src/rxcpp/subjects/rx-synchronize.hpp#L91-L92
`schedule` should be thread-safe by contract if I am not mistaken.

### LINKED ISSUE #492: new_thread scheduler thread safety
Thread sanitizer detects a data race in the following code:
```C++
rxcpp::composite_subscription lifetime;
auto worker =
    rxcpp::observe_on_new_thread().create_coordinator(lifetime).get_worker();
std::vector<std::thread> threads;
for (int i = 0; i < 20; ++i) {
  threads.emplace_back([&worker] {
    worker.schedule([&worker](const auto &) {
      for (size_t i = 0; i < 4294967292l / 6; ++i)
        ;
      printf("Work\n");
    });
  });
}
for (auto &t : threads) {
  t.join();
}
lifetime.unsubscribe();
```
https://gist.github.com/lebdron/883aa78504bc8b9b564e744e2951fb52

Should this bool be made atomic, or there is some other issue with the scheduler?
https://github.com/ReactiveX/RxCpp/blob/ce5dfc2fdd49693cd109a5f5470b73e982c33f35/Rx/v2/src/rxcpp/rx-scheduler.hpp#L97

lebdron:
Closes #491 
Closes #492

### DIFF
--- a/Rx/v2/src/rxcpp/rx-scheduler.hpp
+++ b/Rx/v2/src/rxcpp/rx-scheduler.hpp
@@ -62,12 +62,12 @@ public:
 /// the action uses recurse to coordinate the scheduler and the function.
 class recurse
 {
-    bool& isallowed;
+    std::atomic<bool>& isallowed;
     mutable bool isrequested;
     recursed requestor;
     recurse operator=(const recurse&);
 public:
-    explicit recurse(bool& a)
+    explicit recurse(std::atomic<bool>& a)
         : isallowed(a)
         , isrequested(true)
         , requestor(isrequested)
@@ -94,7 +94,7 @@ public:
 /// recursion is used by the scheduler to signal to each action whether tail recursion is allowed.
 class recursion
 {
-    mutable bool isallowed;
+    mutable std::atomic<bool> isallowed;
     recurse recursor;
     recursion operator=(const recursion&);
 public:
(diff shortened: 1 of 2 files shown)


kirkshoop:
Thank you!

PR #509, opened 2019-08-03
make find_package(rxcpp CONFIG) work through exporting TARGETS

Exctues:
Signed-off-by: exctues <alexnamecode@gmail.com>

The idea is to allow Config mode(not only Module) in find_package function.
So, now in order to import rxcpp in the other project what you need to do is:
1) install rxcpp by "make install"
2) add find_package(rxcpp CONFIG) into CMakeLists.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,23 @@ git clone --recursive https://github.com/ReactiveX/RxCpp.git
 cd RxCpp
 ```
 
+# Installing
+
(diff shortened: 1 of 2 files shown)


kirkshoop:
Nice! Would you please add this to the Readme?

Exctues:
> Nice! Would you please add this to the Readme?

Sure. Can you check it?

kirkshoop:
Thanks!

PR #510, opened 2019-09-10
Add instructions to install RxCpp using vcpkg

vicroms:
Vcpkg is a C++ library manager that can install RxCpp using a single-line command, and can also provide support to integrate RxCpp in CMake projects easily.


kirkshoop:
Thank you!

PR #511, opened 2019-09-25
Observable forward declaration header

lebdron:
Using https://github.com/aras-p/ClangBuildAnalyzer allowed us to check that including `rxcpp/rx.h` was taking significant part of build time. It was possible to reduce it by using `rx-lite`, but the forward declaring header allowed to reduce it even more, because `observable` is used in interfaces.


kirkshoop:
Thank you!

Issue #501, opened 2019-06-04
Throttle missing


wagcampbell:
The current implementation is missing the throttle operator.

The closest I could find is the debounce operator. However, one difference is that the debounce operator will emit the last item in a series of items while throttle will emit the first.

Is there an equivalent operator or combination that will produce the same result?

kirkshoop:
yes, throttle should be added.

it may be possible to use `window_trigger`, `first`, `debounce` to get the first item instead of the last. something like this:

```cpp
auto throttle = [](auto time){
  return [time](auto in){
    auto s = in.publish().ref_count();
    return s.
      window_trigger(s.debounce(time)).
      transform([](auto w){return w.first();})).
      switch_on_next();
  };
};
```

shtaif:
Very nice!

I wonder how one could make such custom `throttle` operator pipable as the other built-in operators?
```cpp
auto sub = rxcpp::observable<>::interval(std::chrono::steady_clock::now(), std::chrono::milliseconds(100), rxcpp::schedulers::observe_on_new_thread())
    | throttle(std::chrono::milliseconds(400))
    | subscribe<long>([](...) { /* ... */ });
```

Thanks! :)

PR #515, opened 2019-10-29
Add throttle operator

cmd-johnson:
I'm used to RxJS and loved the concept, so I wanted to use it in C++ as well. Seeing #501, I thought this would be a good opportunity for me to get a little deeper into understanding the implementation of RxCpp.

This is mostly a copy&paste of the `debounce` operator implementation, edited to implement `throttle` semantics ([ref. RxJS](https://rxjs-dev.firebaseapp.com/api/operators/throttleTime)).

The `throttle` operator re-emits the first value emitted by the source, then swallows all following emissions until the specified timeout has occurred. After that, the next value emitted by the source is emitted again and the cycle continues.

The RxJS implementation also supports emitting the last value emitted by the source when the timeout completes. If this is a feature you would like to see, I can add support for that as well.

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-throttle.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-throttle.hpp
@@ -1,0 +1,271 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+#pragma once
+
+/*! \file rx-throttle.hpp
+
+    \brief  Return an observable that emits a value from the source and then ignores any following items until a particular timespan has passed before emitting another value.
+
+    \tparam Duration      the type of the time interval
+    \tparam Coordination  the type of the scheduler
+
+    \param period        the period of time to suppress any emitted items after the first emission
+    \param coordination  the scheduler to manage timeout for each event
+
+    \return  Observable that emits a value from the source and then ignores any following items until a particular timespan has passed before emitting another value.
+
+    \sample
+    \snippet throttle.cpp throttle sample
+    \snippet output.txt throttle sample
+*/
+
+#if !defined(RXCPP_OPERATORS_RX_THROTTLE_HPP)
+#define RXCPP_OPERATORS_RX_THROTTLE_HPP
+
+#include "../rx-includes.hpp"
+
+#include <iostream>
+
+namespace rxcpp {
+
+namespace operators {
+
+namespace detail {
+
+template<class... AN>
+struct throttle_invalid_arguments {};
+
+template<class... AN>
+struct throttle_invalid : public rxo::operator_base<throttle_invalid_arguments<AN...>> {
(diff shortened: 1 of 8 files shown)


cmd-johnson:
Hmmm, not quite sure why the Travis build fails. It's [only one](https://travis-ci.org/ReactiveX/RxCpp/jobs/604480678?utm_medium=notification&utm_source=github_status) and it fails to install `g++-8`, which all other runners have had no problems with.

Rx/v2/src/rxcpp/operators/rx-throttle.hpp
```diff
@@ -0,0 +1,271 @@
...
+            }
+
+            return std::function<void(const rxsc::schedulable&)>(selectedReset.get());
+        }
+
+        void on_next(T v) const {
+            auto localState = state;
+
+            const auto tp = localState->worker.now().time_since_epoch();
+            std::cout << "on_next(" << v << ") at " << tp.count() / 1000000 << " throttled: " << (localState->throttled ? "true" : "false") << std::endl;
```
kirkshoop: please remove `std::cout` usage

Rx/v2/src/rxcpp/operators/rx-throttle.hpp
```diff
@@ -0,0 +1,271 @@
...
+
+            if (!localState->throttled) {
+                localState->throttled = true;
+
+                state->dest.on_next(v);
+
+                auto work = [v, localState](const rxsc::schedulable&) {
+                    auto produce_time = localState->worker.now() + localState->period;
+
+                    std::cout << "scheduling unthrottle for " << (produce_time.time_since_epoch().count() / 1000000) << std::endl;
```
kirkshoop: please remove `std::cout` usage

Rx/v2/src/rxcpp/operators/rx-throttle.hpp
```diff
@@ -0,0 +1,271 @@
...
+            auto localState = state;
+
+            const auto tp = localState->worker.now().time_since_epoch();
+            std::cout << "on_next(" << v << ") at " << tp.count() / 1000000 << " throttled: " << (localState->throttled ? "true" : "false") << std::endl;
+
+            if (!localState->throttled) {
+                localState->throttled = true;
+
+                state->dest.on_next(v);
+
```
kirkshoop: the call to on_next is not necessarily on the same worker. the following needs to be scheduled on the worker.

```cpp
            if (!localState->throttled) {
                localState->throttled = true;

                state->dest.on_next(v);
            }
```

It would be more efficient to check `throttled` first and then only schedule `state->dest.on_next(v);` when `throttled == false`. To do that throttled would need to be `std::atomic<bool>`.

antekone:
Any news on this PR? Could `throttle` be included in RxCpp?

PR #521, opened 2020-01-05
Changed values_from template parameter to not cast.

tinkerbeast:
The template has several uses as values_from<int, sizeof...(ParamN)>::type(). This causes a cast issue as the second parameter by default is an int whereas we are passing a size_t.


kirkshoop:
thank you!

PR #547, opened 2020-10-07
Included <string> in rx-includes.hpp 

kamleshbhalui:
Included <string> in rx-includes.hpp to compile rxcpp on platform where <string>

is not included indirectly.


kamleshbhalui:
@kirkshoop can you please see this.

Issue #530, opened 2020-02-14
C++20 removed std::result_of


215020267:
**Description:**
In /std:c++17 mode, this triggers a deprecation warning in recent versions of MSVC. In /std:c++latest mode, now that microsoft/STL#380 has been merged, this will trigger an error in VS 2019 16.6 Preview 2.

**Reproduce steps:**
1. git clone --recursive https://github.com/Reactive-Extensions/RxCpp D:\RxCpp\src
2. cd D:\RxCpp\src 
3. set _CL_=/D_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING /D_SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING
4. mkdir build_amd64
5. cd D:\RxCpp\src\build_amd64
6. cmake -G "Visual Studio 15 Win64" -DCMAKE_SYSTEM_VERSION=10.0.17134.0 ..\projects\CMake 2>&1
7. msbuild /m /p:Platform=x64 /p:Configuration=Debug RxCpp.sln /t:Rebuild /p:BuildInParallel=true 2>&1
 
**ErrorMessage:**
F:\gitP\Reactive-Extensions\RxCpp\Rx\v2\src\rxcpp\rx-util.hpp(49): error C2039: 'result_of': is not a member of 'std' [F:\gitP\Reactive-Extensions\RxCpp\build_amd64\test\rxcpp_test_defer.vcxproj]

arolson101:
as a starting point, since result_of is available in versions <= 17, and invoke_result is available in versions >= 17, I did this:

```
// rx-util.hpp
#if __cplusplus >= 201703L
template<class F, class... TN> using result_of_t2 = typename std::invoke_result<F, TN...>::type;
#else
template<class F, class... TN> using result_of_t2 = typename std::result_of<F(TN...)>::type;
#endif
```

then I changed all usages from ```rxu::result_of_t<A(B)>``` to ```rxu::result_of_t2<A, B>```

tcw165:
Do you know the steps to reproduce it on Mac? I'm looking into this issue but couldn't reproduce it.

amelonpie:
Can this issue be closed since patch is merged?

PR #545, opened 2020-09-21
Fix std::result_of removed in C++20

HoshizoraRin:
Fix for #530 based on [comment](https://github.com/ReactiveX/RxCpp/issues/530#issuecomment-648896581) by @arolson101.

I used `__cpp_lib_is_invocable` feature test macro to detect the presence of `std::invoke_result`.

PR #551, opened 2021-01-05
Fix std::result_of removed in C++20

Asaaj:
Fix for #530, intended to be as minimal as possible. Further testing is required, but I wanted to get this in for discussion.


Asaaj:
I wanted a simpler change than #545, which apparently failed. This seems safer to me

Asaaj:
I've never worked with AppVeyor before, but is it reasonably straightforward to get it building with VS2019? Or is RxCpp not claiming to support that version of VS? I can guess what changes are required in appveyor.yml, but I'm not sure the protocol here.

Asaaj:
@kirkshoop with C++20 approaching completion on all major compilers, it would be great to get this fix tested and released soon so this library doesn't gate those who want to upgrade (myself included).

david-alvarez-rosa:
Thanks!!

PR #562, opened 2021-11-22
[Perfomance] Significantly reduce amount of copies/moves inside operators 

victimsnino:
Significantly reduce amount of copies/moves inside operators. Each operator checked and fixed.. In most cases via forwarding/universal reference, in some cases via const reference (if operator is read-only). Each fix covered by unit test

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-debounce.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-debounce.hpp
@@ -126,7 +126,7 @@ struct debounce
                 if(id != state->index)
                     return;
 
-                state->dest.on_next(*state->value);
+                state->dest.on_next(std::move(*state->value));
                 state->value.reset();
             };
 
@@ -140,13 +140,15 @@ struct debounce
             return std::function<void(const rxsc::schedulable&)>(selectedProduce.get());
         }
 
-        void on_next(T v) const {
+        template<typename U>
+        void on_next(U&& v) const {
+            auto vAsShared = std::make_shared<T>(std::forward<U>(v));
             auto localState = state;
-            auto work = [v, localState](const rxsc::schedulable&) {
+            auto work = [vAsShared, localState](const rxsc::schedulable&) {
                 auto new_id = ++localState->index;
                 auto produce_time = localState->worker.now() + localState->period;
 
-                localState->value.reset(v);
+                localState->value.reset(std::move(*vAsShared));
                 localState->worker.schedule(produce_time, produce_item(new_id, localState));
             };
             auto selectedWork = on_exception(
--- a/Rx/v2/examples/linesfrombytes/main.cpp
+++ b/Rx/v2/examples/linesfrombytes/main.cpp
@@ -53,7 +53,7 @@ int main()
                     }) |
                 as_dynamic();
         }) |
-        tap([](vector<uint8_t>& v){
+        tap([](const vector<uint8_t>& v){
             // print input packet of bytes
             copy(v.begin(), v.end(), ostream_iterator<long>(cout, " "));
             cout << endl;
--- a/Rx/v2/src/rxcpp/operators/rx-all.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-all.hpp
@@ -70,7 +70,7 @@ struct all
               done(false)
         {
         }
-        void on_next(source_value_type v) const {
+        void on_next(const source_value_type& v) const {
             auto filtered = on_exception([&]() {
                 return !this->test(v); },
                 dest);
--- a/Rx/v2/src/rxcpp/operators/rx-all.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-all.hpp
@@ -161,12 +161,12 @@ struct member_overload<is_empty_tag>
         class SourceValue = rxu::value_type_t<Observable>,
         class Enabled = rxu::enable_if_all_true_type_t<
             is_observable<Observable>>,
-        class Predicate = std::function<bool(SourceValue)>,
+        class Predicate = std::function<bool(const SourceValue&)>,
         class IsEmpty = rxo::detail::all<SourceValue, rxu::decay_t<Predicate>>,
         class Value = rxu::value_type_t<IsEmpty>>
     static auto member(Observable&& o)
     -> decltype(o.template lift<Value>(IsEmpty(nullptr))) {
-        return  o.template lift<Value>(IsEmpty([](SourceValue) { return false; }));
+        return  o.template lift<Value>(IsEmpty([](const SourceValue&) { return false; }));
     }
 
     template<class... AN>
--- a/Rx/v2/src/rxcpp/operators/rx-amb.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-amb.hpp
@@ -171,8 +171,8 @@ struct amb
                     state->out,
                     innercs,
                 // on_next
-                    [state, st, current_id](value_type ct) {
-                        state->out.on_next(std::move(ct));
+                    [state, st, current_id](auto&& ct) {
+                        state->out.on_next(std::forward<decltype(ct)>(ct));
                         if (!state->firstEmitted) {
                             state->firstEmitted = true;
                             auto do_unsubscribe = [](composite_subscription cs) {
--- a/Rx/v2/src/rxcpp/operators/rx-any.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-any.hpp
@@ -204,12 +204,13 @@ struct member_overload<contains_tag>
         class SourceValue = rxu::value_type_t<Observable>,
         class Enabled = rxu::enable_if_all_true_type_t<
             is_observable<Observable>>,
-        class Predicate = std::function<bool(T)>,
+        class Predicate = std::function<bool(const rxu::decay_t<T>&)>,
         class Any = rxo::detail::any<SourceValue, rxu::decay_t<Predicate>>,
         class Value = rxu::value_type_t<Any>>
     static auto member(Observable&& o, T&& value)
     -> decltype(o.template lift<Value>(Any(nullptr))) {
-        return  o.template lift<Value>(Any([value](T n) { return n == value; }));
+        auto valueAsShared = std::make_shared<rxu::decay_t<T>>(std::forward<T>(value));
+        return  o.template lift<Value>(Any([valueAsShared](const rxu::decay_t<T>& n) { return n == *valueAsShared; }));
     }
 
     template<class... AN>
--- a/Rx/v2/src/rxcpp/operators/rx-any.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-any.hpp
@@ -77,7 +77,7 @@ struct any
               done(false)
         {
         }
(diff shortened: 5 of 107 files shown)


Rx/v2/src/rxcpp/operators/rx-debounce.hpp
```diff
@@ -140,13 +140,15 @@ struct debounce
             return std::function<void(const rxsc::schedulable&)>(selectedProduce.get());
         }
 
-        void on_next(T v) const {
+        template<typename U>
+        void on_next(U&& v) const {
+            auto vAsShared = std::make_shared<T>(std::forward<U>(v));
```
kirkshoop, line 145: an alloc per on_next hurts me :( 
could we just make the lambda mutable and move v instead of copying it?
`localState->value.reset(std::move(v));`

kirkshoop (approved):
Incredible work. Very consistent application of the copy vs. allocation tradeoff. I am not sure that this tradeoff will work for every user, but I would like to try it out. 

The tests are wonderful - they will make this tradeoff selection much easier to maintain.

one small typo is in all the tests

`GIVEN("observale and subscriber")`
observable is missing the second b. Trivial and does not block the PR.

I will try this out on one of my more complicated programs and see if it has any issues.

victimsnino:
Thank you for your time and comments. Waiting for any future comments or merge =)

P.S. I have plans to do something similar for observable part. But I believe, it will be a bit more complicated

kirkshoop:
I tried this out and I cannot compile my project. https://github.com/kirkshoop/twitter

```
/Users/kirk/source/twitter/main.cpp:1115:18: note: candidate function not viable: expects an lvalue for 1st argument
        rxo::map([](Model& m){
                 ^
/Users/kirk/source/twitter/main.cpp:1115:18: note: conversion candidate of type 'model::ViewModel (*)(model::Model &)'
```

The issue seems to be forcing const& value passing. Forwarding would be fine, but const breaks this pattern:

https://github.com/kirkshoop/twitter/blob/master/main.cpp#L1089..L1118

scan owns the Model instance and ViewModel is expected to mutate the instance in scan.

victimsnino:
Is it expected behavior? I mean, from the point of "functional programming" view it is unexpected to change "original" variable in map, isn't it? 

For example, if i do something like this

```cpp
// any caching mechanism
auto observable = rxcpp::observable<>::create<int>([value = 1](rxcpp::subscriber<int> subscriber)
{
   subscriber.on_next(value);
};
// subscriber 1
observable.map([](int& v) {v += 2; return v;}).subscribe();
// subscriber 2
observable.subscribe(); 
```
then subscriber 2 obtains incorrect/unexpected value, isn't it?

victimsnino:
Is it expected behavior? I mean, from the point of "functional programming" view it is unexpected to change "original" variable in map, isn't it?
For example, if i do something like this
```cpp
auto observable = rxcpp::observable<>::just(1).publish();
// subscriber 1
observable.map([](int& v) {v += 2; return v;}).subscribe();
// subscriber 2
observable.subscribe(); 

observable.connect();
```
then subscriber 2 obtains incorrect/unexpected value, isn't it?

Also, your example contains this lines:
```cpp
  start_with(Model{}) |
        rxo::map([](Model& m){
            return ViewModel{m};
        }) |
```

In case of "start_with" ViewModel contains reference to object inside start_with. Depending on implementation it can has as reference to temporal object (if start_with do copy for each new subscriber) or  modify original value inside start_with (if no copy). Not sure if it is expected

kirkshoop:
It is arguable. reduce and scan explicitly mutate the accumulator and scan intentionally emits the accumulator each time it changes. should subsequent operators be allowed to mutate the accumulator? It is safe for scan. the next value will not be accumulated until the stack unwinds from the on_next() that reports the current accumulator.

In practice, I had already made the tradeoff such that Model has a single member that is a shared_ptr. This made it fairly easy to get it to compile with your changes. However, this is a breaking change. 

I think that I will create a new branch to hold the current state of rxcpp before I apply this PR. This will leave a path for others that are dependent on the previous design.

kirkshoop:
The goal would be to allow just to control mutability.

it either can always pass const& to on_next, or take a copy of value before calling on_next or delegate to the caller. Delegation would support passing in a `const int` that would be stored as a const int and passed to on_next as a `const int &`.

The best world (from a C++ perspective) would keep the types directly under the users control - so if the user wanted a value to move or forward or copy or mutable&, through an expression the user would be able to specify the proper type decorations and the operators would preserve those decorations.

That is a lot more work. It was the intent of the original design, but it got lost and ignored too many times and devolved to the inconsistent treatments for values that exist today. If you wanted to take on that task, I would be pleased. I have expectation on you and it will not block this PR.

Thank you!
Kirk

victimsnino:
I would try to think about it after this PR

victimsnino:
@kirkshoop, i have some ideas about providing ability to use mutable values, but i need some clarifications from you:
Current implementation of rxcpp makes copy inside each operator. so, when you do `map([](int& v){...})` actually you obtain reference to local copy (inside map). As a result, yes, you can use `int&` but actually it do nothing in terms of reference (you only avoid copy to pass value inside lambda), but from the meaning of control of value: `reference == copy`.

Suggested there solution has ability to pass real const reference (or rvalue reference) on original object till final observer or any operator (if no any operator with caching mechanism). As a result, each operator operates at each time const lvalue reference or rvalue reference. From this point of view, what does mean "non-const" reference? I mean, non-const reference from const reference -> impossible. non-const reference from rvalue reference - okay, possible, but does it make sense? Looks like you have next options:
1) You don't need any copies and your operator "read-only" -> use const referenece
2) You want to  do some "non-const" operations -> use `int` in lambda and expect copy or move of original object. 
3) You want to do some "non-const reference" operation:
   a) In old implementation it was reference on copy (actually, something like point 2)
   b) If original value was reference -> okay (but it will be rare case even in case of modifying implementation)
   c) If original value was const value, or const lvalue reference, how to deal with it? Looks like we need to provide copy of object inside operator's code. But does it make sense? 

So, final question is: Does it correct, that you want: if original value was some "non-convertible-to-reference" type (const *), then we do copy for passing it to reference argument. If it is true, then ok, i can try to implement it

victimsnino:
If it is applicable, we can do something like this:
https://godbolt.org/z/eGcK9nKf7

TypeDeducer - is special wrapper to resolve all situations. It will be argument in `on_next`. Template parameter - type of argument in lambda.  Several specializations handles all possible variants

call_lambda - imitation of operator (for example, map) where first param - lambda, second - gotten value. Template parameter is equal to  argument of lambda

Check output of execution to understand if it is expected behavior

kirkshoop:
> Current implementation of rxcpp makes copy inside each operator. so, when you do `map([](int& v){...})` actually you obtain reference to local copy (inside map).

I am not sure what you mean..
 
rx-map.hpp
```
        template<class Value>
        void on_next(Value&& v) const
        {
            on_exception_no_return([&]()
                                   {
                                       dest.on_next(this->selector(std::forward<Value>(v)));
                                   },
                                   dest);
        }
```

map() makes no copies of `v`.

kirkshoop:
> 3\. If original value was const value, or const lvalue reference, how to deal with it? Looks like we need to provide copy of object inside operator's code. But does it make sense?

I would expect a compile failure if a source value type was `const T&` and the lambda arg was `T&`. 

I don't like magic :) Conversions should be explicit - either in the definition of the operator (via has to take copies of the value) or in the users code (a lambda taking `T` instead of `T&`/`const T&`

> if original value was some "non-convertible-to-reference" type (const T & or const T), then we do copy for passing it to reference argument.

Thus my answer is that I would like this to fail at compile time. the way to choose the type of the Source Value is to set the type of the Source Value. 

- If the Source Value is `T` or `T&&` then that should be maintained (even after an internal copy, as `via()` does). 
- If the Source Value is `T&` then that should be maintained - __unless__ there is an internal copy - in that case the type passed along should be `T&&` ideally.  
- If the Source Value is `const T&`, then that should be maintained - even if the operator makes a copy internally. __Note:__ it is possible that the correct type to pass along after an internal copy of a `const T&` is `T&&` just like the `T&` case. I am open to either.

The lambda arg type is not going to impact the implementation of any operator. If there is no conversion from the Source Value, then there will be a compile error.

kirkshoop:
> Check output of execution to understand if it is expected behavior.

While that code is cool, it is not the behaviour I would prefer. It is more magical than I like :)

victimsnino:
> > Current implementation of rxcpp makes copy inside each operator. so, when you do `map([](int& v){...})` actually you obtain reference to local copy (inside map).
> 
> I am not sure what you mean..
> 
> rx-map.hpp
> 
> ```
>         template<class Value>
>         void on_next(Value&& v) const
>         {
>             on_exception_no_return([&]()
>                                    {
>                                        dest.on_next(this->selector(std::forward<Value>(v)));
>                                    },
>                                    dest);
>         }
> ```
> 
> map() makes no copies of `v`.

Yes, but `nextdetacher` does:
```cpp
template<class U>
        void operator()(U u) {
            .....
                that->destination.on_error(std::move(ex));
            ....
        }
```

kirkshoop:
This is what I see in nextdetacher

```

template<class U>
        void operator()(U&& u) {
            trace_activity().on_next_enter(*that, u);
            RXCPP_TRY {
                that->destination.on_next(std::forward<U>(u));
                do_unsubscribe = false;
            } RXCPP_CATCH(...) {
                auto ex = rxu::current_exception();
                trace_activity().on_error_enter(*that, ex);
                that->destination.on_error(std::move(ex));
                trace_activity().on_error_return(*that);
            }
        }
```

I do not see a copy..

victimsnino:
> This is what I see in nextdetacher
> 
> ```
> 
> template<class U>
>         void operator()(U&& u) {
>             trace_activity().on_next_enter(*that, u);
>             RXCPP_TRY {
>                 that->destination.on_next(std::forward<U>(u));
>                 do_unsubscribe = false;
>             } RXCPP_CATCH(...) {
>                 auto ex = rxu::current_exception();
>                 trace_activity().on_error_enter(*that, ex);
>                 that->destination.on_error(std::move(ex));
>                 trace_activity().on_error_return(*that);
>             }
>         }
> ```
> 
> I do not see a copy..

With merged changes -> yes =) i told about rxcpp without this merge

kirkshoop:
Ah! Well thank you! That is much better 😳

victimsnino:
I've tried to implement  idea with reference and your requirements above about compilation errors: looks like compilator against this idea and wants to error everytime.

Problem is:
due to now we actually forward values everywhere (instead of providing copies everytime), compilator started to look widely. For example, `on_next` inside map operator now researched for all possible types (const T&, T&& and T& with local changes) So, if i try to pass `map([](int&){}`, then it tries to satisfy ALL possible variants of values for `on_next` inside map operator. As a result, `const int&` and some other can't match for `(int&)` event if it is unused actually (but code exists). Theoretically it should be fixed by templating everything (and compilator would check only really used ways), but `virtual_observer` can be replaced by this way. Any ideas probably???

kirkshoop:
Would you push the new branch as it is so that I can take a look? Perhaps we can move this work into a new issue, do you mind creating an issue for this?

PR #563, opened 2021-11-23
[WIP] Try to provide ability to use mutable/immutable lambdas dynamically

AlexInLog:
To continue thread from #562 .

I've tried to provide override `T&` for on_next. From the point of actual execution it works as expected, but compiler tries to satisfy all possible (and unused) ways of execution (for example, constT& to T&).  Can we solve it?

### DIFF
--- a/Rx/v2/examples/linesfrombytes/main.cpp
+++ b/Rx/v2/examples/linesfrombytes/main.cpp
@@ -53,7 +53,7 @@ int main()
                     }) |
                 as_dynamic();
         }) |
-        tap([](const vector<uint8_t>& v){
+        tap([](vector<uint8_t>& v){
             // print input packet of bytes
             copy(v.begin(), v.end(), ostream_iterator<long>(cout, " "));
             cout << endl;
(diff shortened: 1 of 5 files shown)


victimsnino:
@kirkshoop , moved discussion to this thread as requested

kirkshoop:
One option is to make the virtual function implementations forward to non-virtual internal functions that use enable-if to prune the branches that are invalid and 'magic' the ones that are invalid - `const T&` -> `T&` would make a copy, but would prefer a `T&&` overload of the destination on_next over a `T&` overload.

Another approach is to make the exact type of T supplied to the virtual_observer the only virtual overload for on_next and fail at compile time if the observer passed to the virtual_observer<T> cannot call on_next with T.

I prefer the second option.

victimsnino:
To clarify idea:
Do you want to make type deduction/usage more strict? for example,
```cpp
auto root = rxcpp::observable<>::just(1); // observable<int>;
auto modified = root.map([](int& v){v += 1; return v;}); // observable<int>. compilable due to int convetible to int& . Modify original value inside "just" in this case or just sends copy everytime?
auto ref_modified = modified.map([](int& v)-> int& {return v;}); // observable<int&>

ref_modified.map([](int v){}); // copy of int&
auto const_ref_modified = modified.map([](int& v)-> const int& {return v;}); // observable<int&>
const_ref_modified.map([](int& v){}); // non compilable due to const int& non-convertible to int& 
```

and etc?

kirkshoop:
> Do you want to make type deduction/usage more strict?

@victimsnino, yep, those examples look good to me

PR #564, opened 2021-11-30
[Perfomance] Significantly reduce amount of copies during source operators

victimsnino:
Reduce amount of copies in source operators. Especially for iterate/just/from: reduce from 12-15 copies + 5-6 moves per object + till 2-3 copies OR moves


victimsnino:
@kirkshoop , can you check it?
+ looks like appveyor broken after branch renaming

kirkshoop (approved):

PR #565, opened 2021-11-30
[WIP] Try to improve member declaration via providing all possible overloads as is

AlexInLog:
Attempt to pass operator overloadings to observable as is. As a  result, you can understand that something wrong with your parameters BEFORE compilation due to intellisence and other IDE tricks. No any static asserts, `map_invalid_argument_t` and etc, only SFINAE. Also it improves coding experience in IDE due to providing possible arguments:

image
image

Instead of 

image

I understand your original idea about decoupling operators from observable. In this case you don't need header with implementation: only declaration of functions.

@kirkshoop , what do you think about it?


kirkshoop:
I scanned this and I will need to spend more time contrasting this with the current mechanism. 

If you have time would you try to demonstrate the code change from the current approach to this proposal (this is a separate file so a diff is a little more work).

victimsnino:
@kirkshoop , sorry, I didn’t catch you. Do you mean to try to move all operators to this solution?

Issue #570, opened 2022-01-22
-Werror for unsafe nullptr - `decltype((*(input_type*)nullptr).out((*(Subscriber*)nullptr)))' evaluation


RalphSteinhagen:
First, thanks for this very interesting and well-thought-through library. :+1:

The [shared.cmake](https://github.com/ReactiveX/RxCpp/blob/main/projects/CMake/shared.cmake#L13) prescribes strict type checking via `-Werror` for both [gcc](https://github.com/ReactiveX/RxCpp/blob/677212214755dd282cc71c2263de87e1371f9293/projects/CMake/shared.cmake#L26) and [clang](https://github.com/ReactiveX/RxCpp/blob/677212214755dd282cc71c2263de87e1371f9293/projects/CMake/shared.cmake#L13). 
This fails for at least gcc 11.2.1, clang 13.0.0, and possibly earlier versions for a `decltype(..)` evaluation relying on an `nullptr` cast:

```
[..]/RxCpp/Rx/v2/examples/cep/main.cpp:34:12:   required from here
[..]/RxCpp/Rx/v2/src/rxcpp/rx-coordination.hpp:62:56: error: ‘this’ pointer is null [-Werror=nonnull]
   62 |         using type=decltype((*(input_type*)nullptr).out((*(Subscriber*)nullptr)));
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```

Maybe this pointer/type magic could be replaced by a safer alternative. 

Since this library aims at newer compiler/C++ standards, maybe the various typedefs could possibly be replaced by `using type=...`.

RalphSteinhagen:
@kirkshoop or anybody else: any chance to get feedback on this and the corresponding PR? 

We liked to use it in our C++20 project but the compiler warnings turned into errors would be a substantial technical burden.

Thanks in advance and much appreciated! :+1:

RalphSteinhagen:
@kirkshoop thanks for merging and accepting this! Issue is fixed by PR closing now.

PR #572, opened 2022-01-28
replace C-style pointer cast type deduction by 'std::declval<>()'

RalphSteinhagen:
First of all: thanks for RxCpp and the great work you put into it! 
It's a really neat and useful library and I wonder why it isn't already part of the C++ standard! :+1:

This PR primarily addresses issue #570 and replaces pointer-type casting which causes warnings and (w/ `-Werror`) errors on newer compilers.

While refactoring, I noticed that some structures appear rather old (pre C++11) and could be simplified which I hope may help readability, modernizing towards C++20 (if this isn't excluded), and help new developers. Some of the low-hanging fruits (w/o breaking the apparent <=C++14 constraint) were:
  * updated Catch2 to v2.13.8 (216713a) - the old version isn't compatible anymore with more modern OSes.
  * replacing C-style typedef type aliasing by 'using type = ' see also [C++ Core Guidelines - T.43](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using)
N.B. this should make later refactoring (and possible introduction of 'concepts') a bit easier ...
  * replaced `std::is_same<...>::value` by `std::is_same_v<...>`
  * replaced `std::conditional<...>::type` by `std::conditional_t<...>`

I didn't tackle the ubiquitous `std::enable_if<...>::type` by `std::enable_if_t<...>` simplification and swapping of confusing ternary statements because of the sheer number of lines that would be impacted and because many of these lines wouldn't probably be needed anymore if this library would adopt C++20 and concepts (see also [C++ Core Guidelines - T.48](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t48-if-your-compiler-does-not-support-concepts-fake-them-with-enable_if)).

Let me know what you think and hope this fits the coding standard and style this library aims for.

### DIFF
--- a/.github/workflows/rxcpp-ci.yml
+++ b/.github/workflows/rxcpp-ci.yml
@@ -77,7 +77,7 @@ jobs:
             os: windows-latest,
             build_type: Debug,
             cc: "cl", cxx: "cl",
-            environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
+            environment_script: "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=17",
             experimental: true
           }
@@ -86,7 +86,7 @@ jobs:
             os: windows-latest,
             build_type: RelWithDebInfo,
             cc: "cl", cxx: "cl",
-            environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
+            environment_script: "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=17",
             experimental: true
           }
--- a/Rx/v2/src/rxcpp/operators/rx-all.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-all.hpp
@@ -42,11 +42,11 @@ using all_invalid_t = typename all_invalid<AN...>::type;
 template<class T, class Predicate>
 struct all
 {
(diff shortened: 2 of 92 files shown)


kirkshoop (approved):

guhwanbae:
@RalphSteinhagen , @kirkshoop 
A while ago, I found that C++ standard version is not specified in the CMakeLists.txt file for unit tests. So there unit test are not compiled with gcc-9. gcc-9 use C++14 as default standard version.

So, Is C++17 minimum standard version of this project now?

kirkshoop:
@guhwanbae - no, I do not intend rxcpp to drop cpp14 support yet. Did I merge something that requires c++17?

guhwanbae:
@kirkshoop , Yes, 
https://github.com/ReactiveX/RxCpp/pull/572/commits/79addf123c6b94b8bda85099ae96b4f69919247b
It replaced std::is_same::value with std::is_same_v alias since C++17.

RalphSteinhagen:
@guhwanbae @kirkshoop I am terribly sorry. It was not my intention to implicitly boost to C++17.
I relied on the existing cmake definitions and CI/CD definitions which neither reported errors nor warnings and it thus slipped through, sorry.

Since this is a very simple alias definition and in order to move forward, one could
  a) provide an alias (checking for the GCC/clang capabilities) and define an `std::is_same_v<...>` if <C++17, or
  b) search-and-replace each `std::is_same_v<...>` occurrence back to ` `std::is_same<...>::value`.
  
 The first is a bit dirty if RxCpp is targeted to remain at C++14 compatibility for the foreseeable future, and the second -- well a bit verbose. 

 @kirkshoop In any case, I'd be happy to make a PR to address this issue either way... just let me know your preference.

kirkshoop:
No worries. I was also trusting the ci to cetch these types of issues. 

I would prefer a local definition of rxcpp::is_same_v with a check for the std version macro for std::is_same_v. If the library has it alias it, if it does not, define it. 

I would also like to have the ci enforcement strengthened so that it catches these things. 

Thanks!

PR #581, opened 2022-03-14
enforce C++14 in cmake + added rxcpp::is_same_v C++14 compatible definition ...

RalphSteinhagen:
... in view of >=C++17's `std::is_same_v`

see also post-PR discussion at: https://github.com/ReactiveX/RxCpp/pull/572#issuecomment-1066465190-permalink

I tried to keep it as non-intrusive as possible and defined a `rxcpp::is_same_v` template pair in `rx-trace.cpp` since this seems to be the place that most (all?) of the other includes reference. Needed the `static constexpr bool ...` because C++14 does not allow for 'inline' and since `is_same_v` is a value and not type trait*.

@guhwanbae @kirkshoop hope and please let me know if this helps with the issue mentioned above.

P.S. *I noticed that the build times are quite large (possibly because of SFINAE, enable_ifs, etc.). Would be worthwhile to perhaps consider moving RxCpp to the current standard  (C++20)??  ... more features (concepts), lots of goodies, ranges (that seem to be similar/related to RxCpp), simpler/more succinct code, and faster compile-times. Food for thought.

### DIFF
--- a/Rx/v2/src/rxcpp/operators/rx-combine_latest.hpp
+++ b/Rx/v2/src/rxcpp/operators/rx-combine_latest.hpp
@@ -68,7 +68,7 @@ struct is_combine_latest_selector_check {
 
     using type = decltype(check<selector_type, rxu::decay_t<ObservableN>...>(0));
 
-    static const bool value = !std::is_same_v<type, tag_not_valid>;
+    static const bool value = !rxcpp::is_same_v<type, tag_not_valid>;
 };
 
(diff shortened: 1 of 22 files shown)


RalphSteinhagen:
@kirkshoop @guhwanbae is there an issue with this PR? It has been open for quite some time.

There is already a huge body of implementations available and we are evaluating whether to use RxCpp in other applications (mostly C++20). Having fewer compiler warnings/cleaner code wouldn't harm.

PR #588, opened 2022-09-04
update README.md

coderSHeE:
I make some addition of adding av comment in line 38 , and also add some spaces between the main function , 
also removed some extra elements in line 72  & 74 ,
more added one end comment in line 122 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ using namespace Rx;
 using namespace std;
 using namespace std::chrono;
 
+// main section begin
+
 int main()
 {
     random_device rd;   // non-deterministic generator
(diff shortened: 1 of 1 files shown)


coderSHeE:
I make some addition of adding av comment in line 38 , and also add some spaces between the main function ,
also removed some extra elements in line 72 & 74 ,
more added one end comment in line 122 .

coderSHeE:
I make some addition of adding av comment in line 38 , and also add some spaces between the main function ,
also removed some extra elements in line 72 & 74 ,
more added one end comment in line 122 .

PR #606, opened 2024-07-30
Fix gcc 14 compilation

### LINKED ISSUE #605: Library fails to compile with gcc14
Compiling a simple program which includes rx-includes.hpp fails with gcc 14.

```c++
#include <iostream>
#include "rxcpp/rx-includes.hpp"

int main() {
    std::cout << "wow" << std::endl;
    return 0;
}
```

The issue appears to be that the assignment operators of ~~`on_next_notification`~~ and `on_error_notification` attempt to modify a private const field ~~[here](https://github.com/ReactiveX/RxCpp/blob/761b932a80e2be6e2b62d232e754bd96fc448946/Rx/v2/src/rxcpp/rx-notification.hpp#L130)~~ and [here](https://github.com/ReactiveX/RxCpp/blob/761b932a80e2be6e2b62d232e754bd96fc448946/Rx/v2/src/rxcpp/rx-notification.hpp#L158), respectively. Apparently this compiles in older versions of gcc because it was simply not attempting to check assignments inside of templates, unless the templates are being used. With [this change](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=18474) in gcc 14 however, simple assignments are now checked even inside of templates. This would be my guess as to why this code now fails to compile.

The fix seems to be to simply delete the assignment operator for ~~`on_next_notification`~~ and `on_error_notification`, as the two structs are only ever instantiated as `shared_ptr` so it is not needed. In any case, compilation will fail with older versions of gcc if a user writes code which causes a variant of the assignment operator template to be generated.

Edit: The project I've been working on uses RxCpp v4.1.1, but apparently the const declaration of the `T value` field of `on_next_notification` has been removed on master.

colbysparks:
Cause of bug and proposed change outlined in #605.

closes #605

### DIFF
--- a/Rx/v2/src/rxcpp/rx-notification.hpp
+++ b/Rx/v2/src/rxcpp/rx-notification.hpp
@@ -155,7 +155,7 @@ private:
         }
         on_error_notification(const on_error_notification& o) : ep(o.ep) {}
         on_error_notification(const on_error_notification&& o) : ep(std::move(o.ep)) {}
-        on_error_notification& operator=(on_error_notification o) { ep = std::move(o.ep); return *this; }
+        on_error_notification& operator=(on_error_notification o) RXCPP_DELETE;
         void out(std::ostream& os) const override {
             os << "on_error(";
             os << rxu::what(ep);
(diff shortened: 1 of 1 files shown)


keef-cognitiv:
Crickets :cry:

